PR #2, opened 2020-10-30
Use HTTPS

hpdeifel:
especially since the admin token is so sensitive

### DIFF
--- a/synadm.py
+++ b/synadm.py
@@ -54,7 +54,7 @@ class Synapse_admin (object):
         headers={'Accept': 'application/json' }
         tokenpart=f"access_token={self.token}"
         # take care of putting & or ? at end of urlpart in calling method already!
-        url="http://{}:{}/_synapse/admin/{}{}".format(self.host, self.port,
+        url="https://{}:{}/_synapse/admin/{}{}".format(self.host, self.port,
               urlpart, tokenpart)
         log.debug('_get_synapse url: {}\n'.format(url))
         try:


JOJ0 (commented):
hi hpdeifel, thanks a lot for the pull-request. Yes https was on the todolist already but I was unsure if it will work that easy and is the right decision.

Did you test this? What about the certificate? I am not sure wether the certificate on 8448 is self-signed or an official one signed by matrix.org CA or whatever.

On my personal homeserver I don't have 8448 running at all. 8008 is _only_ served to localhost and furthermore I have an apache configuration that serves 8008 via https port 443 to the outside world. I am not quite sure if in my case the serving to matrix clients _and_ also the serving to federated matrix servers is done on the same port 443 but I would assume so. I would need to look this up and check my configuration again....

Also as far as I remember having 8008 served to localhost **only** and then use a seperate "real" webserver configured as reverse-proxy is the recommended way by matrix...... had to ook this up again and yes, this is what they recommend: https://github.com/matrix-org/synapse/blob/develop/docs/reverse_proxy.md

hpdeifel:
Yeah, well. Wasn't thinking about the case where the tool runs on localhost. I expose the admin API to the internet and the absolutely *need* to use https here.

I think the best way forward (and the one used in my old tool) is to let the user specify the whole base-url of synapse (or even the `admin` endpoint directly). That way, people can use https/http depending on their use-case and people where synapse listens on 443 don't even have to specify the port

JOJ0:
> Yeah, well. Wasn't thinking about the case where the tool runs on localhost. I expose the admin API to the internet and absolutely _need_ to use https here.

in that case: sure as hell that's the only way haha :-)) matrix does not recommend exposing _synapse/admin to the internet at all. I need to check wether I did expose this. I think I dont and just _synapse/client is serverd to the world........just checked. this is how i did it. only matrix-client are allowed and _synapse/admin is not even served to the outside:


```
        ProxyPreserveHost On
        AllowEncodedSlashes NoDecode
        ProxyPass /_matrix http://127.0.0.1:8008/_matrix nocanon
        ProxyPassReverse /_matrix http://127.0.0.1:8008/_matrix
```

>
> I think the best way forward (and the one used in my old tool) is to let the user specify the whole base-url of synapse (or even the `admin` endpoint directly). That way, people can use https/http depending on their use-case and people where synapse listens on 443 don't even have to specify the port

yes, you are absolutely right it definitely should be configurable if a user choses http or https, not sure how i will realize it but will change that in the configuration code soon.

but still: did you test the tool against https://localhost:8448/_synapse/admin ? Does python requests module not complain about the certificate? You use a self-signed one or the one which was shipped with the synapse-package?

hpdeifel:
> but still: did you test the tool against https://localhost:8448/_synapse/admin ? Does python requests module not complain about the certificate? You use a self-signed one or the one which was shipped with the synapse-package?

That won't work, except if you really create a self-signed cert for the domain `localhost`. I didn't test it on the server directly, but it's clear that it wouldn't work.

My use-case really is remote usage of the admin API, which means I couldn't test the tool without https and bearer-auth. One case where to expose the admin API is if the synapse administrator doesn't have shell access to the server, which happens e.g. if the hosting is provided by a third party.

JOJ0:
> > but still: did you test the tool against https://localhost:8448/_synapse/admin ? Does python requests module not complain about the certificate? You use a self-signed one or the one which was shipped with the synapse-package?
> 
> That won't work, except if you really create a self-signed cert for the domain `localhost`. I didn't test it on the server directly, but it's clear that it wouldn't work.

Yes certainly, I didn't think it through entirely. Of course there needs to be a proper FQDN and fitting certificate for ssl working as designed. My thoughts just went as far as: If I comment-in the provided stanzas for 8448 in a default homerserver.yaml, is there any certificate present already? What is it? How is this supposed to work.... Anyway, nevermind! :-)

> My use-case really is remote usage of the admin API, which means I couldn't test the tool without https and bearer-auth. One case where to expose the admin API is if the synapse administrator doesn't have shell access to the server, which happens e.g. if the hosting is provided by a third party.

ssl enpoints should now be supported: https://github.com/JOJ0/synadm/commit/6a06ad41c30b3fe5e1aaa3860533892d27113902

I am closing this one.

PR #3, opened 2020-10-30
Use header instead of query param for auth

hpdeifel:
This avoids leaking the access token in HTTP server logs and is explicitly
recommended by the matrix specification.

(depends on #2)

### DIFF
--- a/synadm.py
+++ b/synadm.py
@@ -51,11 +51,10 @@ class Synapse_admin (object):
         self.token = token
 
     def _get(self, urlpart):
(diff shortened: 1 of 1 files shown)


JOJ0:
please remove the https commit for now, so i can accept the pull right now. I will deal with https or not configuration later on as discussed in https://github.com/JOJ0/synadm/pull/2

hpdeifel:
fixed

JOJ0:
sending the token not in the url is a very valuable not to say essential change! :-) thanks a lot!

PR #6, opened 2020-12-13
Split the code into modules and some more refactoring

kaiyou:
After the `axon` project was discussed on the TWIM Matrix channel, @JOJ0 contacted me so we would merge our projects. This MR is a first attempt at bringing the interesting parts of `axon` and making sure that `synadm` is a fit for my use cases as well, so I can keep using it and contribute to it.

Main changes include : 

 - split the synadm module into multiple files and make it a package
 - simplify the exception handling everywhere (stop catching individual exception if we do not handle them individually)
 - simplify the HTTP api code
 - stop passing a complex dictionary in the click context and use a helper class instead
 - improve pep8 and pylint compliance (documentation, line handling, quotes, imports, etc.)


JOJ0:
Hi @kaiyou, thanks a lot for your latest commits. I did some testing with those: Some of the partly broken things are fixed now and think I already found the root-cause of some more issues. I might leave some notes about those findings right next to the code.

As already discussed in our Matrix chat room I will try to try to intensively test everything in then next couple of days, merge this into master and fix the rest there.

So for other people wondering about this PR not being merged yet: There is big changes of synadm's internals in here, which meant a lot of testing, discussion and planning ahead so synadm's development can go into a direction @kaiyou and me are both happy with. I think we are on track now and should be done with these first steps soon.

synadm/cli/__init__.py
```diff
@@ -132,13 +136,16 @@ def output(self, data):
     default="~/.config/synadm.yaml",
     help="configuration file path", show_default=True)
 @click.pass_context
-def root(ctx, verbose, output, config_file):
+def root(ctx, verbose, batch, output, config_file):
     """ Synapse Administration toolkit
     """
-    ctx.obj = APIHelper(config_file, verbose, output)
+    ctx.obj = APIHelper(config_file, verbose, batch, output)
     if ctx.invoked_subcommand != "config" and not ctx.obj.load():
```
JOJ0, line 147: Seems like ctx.obj.load() is not executed here and thus calling `synadm config` does not read from config file.
Something like `helper_load_success = ctx.obj.load()` and then asking for this variable in the condition fixes it.

JOJ0 (approved):

synadm/cli/__init__.py
```diff
@@ -172,35 +161,30 @@ def root(ctx, verbose, output, config_file):
...
+        "token": click.prompt("Synapse admin user token",
+            default=api.config.get("token", token)),
+        "base_url": click.prompt("Synapse base URL",
+            default=api.config.get("base_url", base_url)),
+        "admin_api_path": click.prompt("Synapse admin API path",
+            default=api.config.get("admin_api_path", admin_api_path)),
+        "format": click.prompt("How should data be viewed by default?",
+            default=api.config.get("format", output),
+            type=click.Choice(['human', 'json', 'yaml', 'pprint']))
+    })
```
JOJ0, line 203: After writing new config to file it needs to be read again here otherwise if config was called automatically because a user issued any command with an incomplete /broken config file, after going through the configurator it would fail. This fixes it: `helper.config.load()'

synadm/cli/__init__.py
JOJ0: Also after writing new config to file it needs to be read again here otherwise if config was called automatically because a user issued any command with an incomplete /broken config file, after going through the configurator it would fail. Putting this after the write fixes it: `helper.config.load()`

synadm/cli/__init__.py
```diff
@@ -41,80 +41,75 @@ class APIClient(api.Synapse_admin):
     }
 
     CONFIG = {
-        "user": None,
-        "token": None,
-        "base_url": None,
+        "user": "",
+        "token": "",
+        "base_url": "",
```
JOJ0, line 47: admin_path should be admin_api_path so it reflects the name of the `synadm config` option

synadm/cli/__init__.py
```diff
@@ -41,80 +41,75 @@ class APIClient(api.Synapse_admin):
...
-                log.error(f"config entry {key} missing")
-                log.error("please run: " + sys.argv[0] + " config")
-                raise SystemExit(2)
+            if not value:
+                self.log.error(f"config entry {key} missing")
+                return False
+        self.api = api.Synapse_admin(
+            self.log,
+            self.config["user"], self.config["token"],
+            self.config["base_url"], self.config["admin_path"]
```
JOJ0, line 105: Also here the config dict's key needs to be admin_api_path

JOJ0:
A couple of days became two weeks but finally merging this into master now. Will push some fixes to master afterwards.

PR #15, opened 2021-02-05
[WIP] Implement command for make_room_api

rht:
Currently not working. I kept getting this error
```
ERROR HTTPError while querying Synapse: 400 Client Error: Bad Request for url: https://matrix.domain.com/_synapse/admin/v1/rooms/!RoOmId:matrix.domain.com/make_room_admin
```

not sure what the cause is.

(I'm `@landau` in #synapse:matrix.org)

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -153,7 +153,9 @@ class SynapseAdmin:
     def room_make_admin(self, room_id, user_id):
         """ Make a user into an admin of a room
         """
-        data = {"user_id": user_id}
+        data = {}
+        if user_id:
+            data.update({"user_id": user_id})
         return self.query("post", f"v1/rooms/{room_id}/make_room_admin", data=data)
 
     def room_media_list(self, room_id):
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -151,7 +151,8 @@ class SynapseAdmin:
         return self.query("post", f"v1/rooms/{room_id}/delete", data=data)
 
     def room_make_admin(self, room_id, user_id):
-        """ Make a user into an admin of a room
+        """ Grant a user room admin permission. If the user is not in the room,
+        and it is not publicly joinable, then invite the user.
         """
         data = {}
         if user_id:
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -150,6 +150,12 @@ class SynapseAdmin:
             data.update({"message": message})
         return self.query("post", f"v1/rooms/{room_id}/delete", data=data)
 
+    def room_make_admin(self, room_id, user_id):
+        """ Make a user into an admin of a room
+        """
+        data = {"user_id": user_id}
+        return self.query("post", f"v1/rooms/{room_id}/make_room_admin", data=data)
+
     def room_media_list(self, room_id):
         """ Get a list of known media in an (unencrypted) room.
         """
--- a/synadm/cli/room.py
+++ b/synadm/cli/room.py
@@ -163,7 +163,10 @@ def search_room_cmd(ctx, search_term, from_, limit):
 
 @room.command()
 @click.argument("room_id", type=str)
-@click.argument("user_id", type=str)
+@click.option(
+    "--user-id", "-u", type=str,
+    help="""By default the server admin (the caller) is granted power, but
+    another user can optionally be specified.""")
 @click.pass_obj
 def make_admin(helper, room_id, user_id):
     """ Make a user into an admin of a room
--- a/synadm/cli/room.py
+++ b/synadm/cli/room.py
@@ -159,3 +159,14 @@ def search_room_cmd(ctx, search_term, from_, limit):
                .format(search_term.capitalize()))
     ctx.invoke(list_room_cmd, from_=from_, limit=limit,
                name=search_term.capitalize())
+
+
+@room.command()
+@click.argument("room_id", type=str)
+@click.argument("user_id", type=str)
+@click.pass_obj
+def make_admin(helper, room_id, user_id):
+    """ Make a user into an admin of a room
+    """
+    out = helper.api.room_make_admin(room_id, user_id)
+    helper.output(out)
--- a/README.md
+++ b/README.md
@@ -175,6 +175,7 @@ python3 setup.py install
   * [x] `room members <room id>`
   * [x] `room delete <room id>`
   * [ ] Additional commands derived from rooms API's
+    * [x] `room make-admin <room id> <user id>`
     * [ ] `room count`
     * [ ] `room top-complexity`
     * [ ] `room top-members`


rht:
Note that other `synadmin room` commands work perfectly.

rht:
Also, not yet implemented that according to https://github.com/matrix-org/synapse/blob/master/docs/admin_api/rooms.md#make-room-admin-api
> By default the server admin (the caller) is granted power, but another user can optionally be specified

so the `user_id` argument should be optional.

JOJ0:
> so the `user_id` argument should be optional.

use an option decorator instead of argument

JOJ0:
many thanks for the PR, actually looks correct to me, will test it and report back tonight or tomorrow.

JOJ0:
Same here, but I don't see the problem. I even checked with requests debugging enabled. Looks correct. Maybe Synapse doesn't support this API anymore?

```
$ synadm -v room make-admin '!mMozyfiOAeDNPhtKWc:peek-a-boo.at' @testuser1:peek-a-boo.at
INFO  Querying post on http://localhost:8008/_synapse/admin/v1/rooms/!mMozyfiOAeDNPhtKWc:peek-a-boo.at/make_room_admin
send: b'POST /_synapse/admin/v1/rooms/!mMozyfiOAeDNPhtKWc:peek-a-boo.at/make_room_admin HTTP/1.1\r\nHost: localhost:8008\r\nUser-Agent: python-requests/2.25.0\r\nAccept-Encoding: gzip, deflate\r\nAccept: application/json\r\nConnection: keep-alive\r\nAuthorization: Bearer secretxtokenxxxxxxxxxxxx\r\nContent-Length: 39\r\nContent-Type: application/json\r\n\r\n'
send: b'{"user_id": "@testuser1:peek-a-boo.at"}'
reply: 'HTTP/1.1 400 Bad Request\r\n'
header: Transfer-Encoding: chunked
header: Server: Synapse/1.24.0
header: Date: Fri, 05 Feb 2021 23:21:26 GMT
header: Content-Type: application/json
header: Cache-Control: no-cache, no-store, must-revalidate
header: Access-Control-Allow-Origin: *
header: Access-Control-Allow-Methods: GET, HEAD, POST, PUT, DELETE, OPTIONS
header: Access-Control-Allow-Headers: Origin, X-Requested-With, Content-Type, Accept, Authorization, Date
ERROR HTTPError while querying Synapse: 400 Client Error: Bad Request for url: http://localhost:8008/_synapse/admin/v1/rooms/!mMozyfiOAeDNPhtKWc:peek-a-boo.at/make_room_admin
null
...
```

rht:
Thanks for the debug. The API still exists. I tried a raw curl and got this: `{"errcode":"M_UNKNOWN","error":"No local admin user in room with power to update power levels."}`.

JOJ0:
please post your curl. the only thing I see is that data is sent with b, which means bytes string, but that's the same in other post calls. I just checked.

rht:
Another evidence that the API still exists: https://github.com/matrix-org/synapse/blob/d882fbca388692f475673007df4f7ac394446b46/synapse/rest/admin/rooms.py#L424.

JOJ0:
I'll postpone further investigation, it's late around here! Thanks so far, we'll figure that out!

rht:
My curl command: `curl -X POST 'https://matrix.domain.com/_synapse/admin/v1/rooms/!roomId:matrix.domain.com/make_room_admin\?access_token=accestoken'` (the json content is optional).

JOJ0:
does the curl work with the user in json content?

JOJ0:
BTW: will make this configurable soon, helps debugging and currently have to hack in manually: https://gist.github.com/Daenyth/b57f8522b388e66fcf3b

rht:
> does the curl work with the user in json content?

Got the same error with without json `{"errcode":"M_UNKNOWN","error":"No local admin user in room with power to update power levels."}`.

JOJ0:
> > does the curl work with the user in json content?
> 
> Got the same error with without json `{"errcode":"M_UNKNOWN","error":"No local admin user in room with power to update power levels."}`.

well, this is at least a proper response of the api, so it seems it is a proper request and understood.

synadm/cli/room.py
```diff
@@ -159,3 +159,14 @@ def search_room_cmd(ctx, search_term, from_, limit):
                .format(search_term.capitalize()))
     ctx.invoke(list_room_cmd, from_=from_, limit=limit,
                name=search_term.capitalize())
+
+
+@room.command()
+@click.argument("room_id", type=str)
+@click.argument("user_id", type=str)
```
JOJ0, line 166: since this is optional, make it an option. hava a look on other commands how to use @click.option decorators. best use --user-id, -u, also add help and describe the behaviour (I often copy paste and shorten stuff from the api documentation ;-)

synadm/api.py
```diff
@@ -147,6 +147,12 @@ def room_delete(self, room_id, new_room_user_id, room_name, message,
             data.update({"message": message})
         return self.query("post", f"v1/rooms/{room_id}/delete", data=data)
 
+    def room_make_admin(self, room_id, user_id):
+        """ Make a user into an admin of a room
+        """
+        data = {"user_id": user_id}
```
JOJ0, line 153: ask for existence of user_id and only add the data dict if it's there, otherwise make data an empty dict

rht:
@JOJ0 do you want me to update the PR to reflect your suggestions first, or debug first?

JOJ0:
> @JOJ0 do you want me to update the PR to reflect your suggestions first, or debug first?

please update this one. I have some corrections in my comments in a minute too (bool vs None)

synadm/cli/room.py
JOJ0: If you use is_flag than click will always send a boolean! So the default value will be true or false and not None! Work with that! Sorry for misleading you here!

JOJ0:
sorry again. my earlier comments where correct already. It's too early and I need to get breakfast I guess. All good! Work away! Thanks a ton in advacne!

JOJ0:
Hi @rht I can reproduce your findings so far: curl is working, I get proper responses as you did but still no luck with getting it through synadm.

I implemented requests debugging. It's in master, merge it into your branch, then user `synadm -vvv ...` to activate.

BTW I am using `Synapse matrix-synapse-py3    1.26.0+bionic1`

rht:
I rebased this PR. Will try with `-vvv`.

JOJ0:
Just another quick idea that might help debugging this: Remove the try except wrapper in SynapseAdmin.query() and see if you get anything more specific. HTH

rht:
I figured out more detail behind the 400 bad request. It is actually consistent with curl.
You only print out `resp.json()` when `resp.ok` is `True`.
When I removed the `resp.ok`, I got:
`{'errcode': 'M_UNKNOWN', 'error': 'No local admin user in room with power to update power levels.'}`

So the code is probably working correctly.

JOJ0:
that's great news! thanks for getting behind this riddle. I'll merge pr #19 now and will have a look on this one soon.

JOJ0:
well, on the quick look the only thing missing here is making user an option instead of mandatory argument and only add it to the api request if necessary.

rht:
I have added a 2nd commit to make `--user-id` optional. I tested by manually printing out `user_id` inside the `room_make_admin` function. Can't test the actual API because I kept getting `{'errcode': 'M_UNKNOWN', 'error': 'No local admin user in room with power to update power levels.'}`

synadm/api.py
```diff
@@ -154,7 +154,9 @@ def room_delete(self, room_id, new_room_user_id, room_name, message,
     def room_make_admin(self, room_id, user_id):
         """ Make a user into an admin of a room
         """
```
JOJ0, line 156: Please change the docstring (which is --help in the end) to:

Grant a user the highest power level available in the room. If the user is not in the room, and it is not publicly joinable, then invite the user.

(I hope I got this right, the sentence from the docs is pretty difficult to read (https://github.com/matrix-org/synapse/blob/master/docs/admin_api/rooms.md#make-room-admin-api) and I am not sure if I understand the meaning correctly, so please review if my sentence makes sense here, thanks!)

rht: Why did you omit "local" from "...highest power available to a local user..."?

JOJ0: i ditched even more from the sentence as I thought it's complicated as it is.

pls make a suggestion i dont get the meaning of it i think

JOJ0:
```
Grant a user room admin permission. If the user is not in the room, and it is not publicly joinable, then invite the user. 
```

The first short sentence is what will be shown in `synadm room -h` (or even without -h), so that's the most important message for the enduser to be seen.

I don't have a better idea for now. On the very end of my two sentences we could additionally add something that is explaining in detail what exactly room admin permission means. But I don't understand it 100% so let's just leave this out for now and get this finally merged :-) Sorry that this PR took quite some time, but I think it was worth it and thanks again for fixing those nasty bugs! Great additions to the project!

rht: Sounds good. Added in 3rd commit.

PR #16, opened 2021-02-05
Format everything with Black

rht:
(https://github.com/psf/black)

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -8,8 +8,7 @@ import requests
 
 
 class SynapseAdmin:
-    """ Synapse API client
-    """
+    """Synapse API client"""
 
     def __init__(self, log, user, token, base_url, admin_path, timeout):
         self.log = log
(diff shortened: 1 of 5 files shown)


JOJ0:
No, please don't! All the code recently was refactored according to PEP8 in PR#6 (using pylint and he might also have used flake8). I think it's fine already. What exactely are your concerns with the current formatting?

synadm/api.py
```diff
@@ -110,33 +114,34 @@ def user_modify(self, user_id, password, display_name, threepid,
...
+        """Get a list of room members"""
         return self.query("get", f"v1/rooms/{room_id}/members")
 
-    def room_delete(self, room_id, new_room_user_id, room_name, message,
-                    block, no_purge):
-        """ Delete a room and purge it if requested
-        """
+    def room_delete(
+        self, room_id, new_room_user_id, room_name, message, block, no_purge
+    ):
```
JOJ0, line 140: For example, this is awful IMHO, the version on the left is properly readable already and spare a line. Using a magic auto-formatting tool is not always the best solution. Still: IMHO, Thanks for the effort anyway :-) I currently like to use recommendations provided by a tool called flake8.

rht:
I think it is whether you like Black or you hate Black. What I did was just to apply Black by default. The point is that `Black` is like `gofmt` where there is only 1 true style to end all the formatting wars. If you haven't heard of it, see https://github.com/psf/black#used-by.

JOJ0:
Yeah it does make sense but I think I am not ready for it at this point. Some things I think get rather longish horizontally. Ask me in a couple of months again. I might change my mind! :-) But definitely thanks anyway! Appreciated!

PR #17, opened 2021-02-06
API: Specify param as long as it is not None

rht:
I have noticed this in other params too. I'm not sure if it is a bug, so I'm making this test PR.

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -73,7 +73,7 @@ class SynapseAdmin:
         """ Set the user password, and logs the user out if requested
         """
         data = {"new_password": password}
-        if no_logout:
+        if no_logout is not None:
             data.update({"logout_devices": no_logout})
         return self.query("post", f"v1/reset_password/{user_id}", data=data)
 


synadm/api.py
```diff
@@ -73,7 +73,7 @@ def user_password(self, user_id, password, no_logout):
         """ Set the user password, and logs the user out if requested
         """
         data = {"new_password": password}
-        if no_logout:
+        if no_logout is not None:
```
JOJ0, line 76: this is exactely the same as "if no_logout" ;-)

JOJ0: Alright I'll give you that. It indeed should be your version according to https://www.python.org/dev/peps/pep-0008/#programming-recommendations

So if you find the time, you could as well extend this PR and fix other occurrences. This pattern is used a lot in api.py

rht: There is a legitimate corner case where the user wants to explicitly specify `False`, but with `if no_logout`, this explicit `False` doesn't get passed in. Things would be fine if the default is `False`, but if the default is `True`, then this becomes unexpected behavior.

JOJ0: well, I should have read my own code better before answering: I just had a look and see that actually the default actually is False:
```
@click.option(
    "--no-logout", "-n", is_flag=True, default=False,
```

Am I right that actually using a boolean comparison as in my original version would be correct to use? (No need for "is not None")

Or am I on the wrong foot? :-)

JOJ0: Ah I see, you are talking about the default of the API iteself being true (https://github.com/matrix-org/synapse/blob/master/docs/admin_api/user_admin_api.rst#reset-password)

JOJ0: ok but since no_logout _always_ comes as a boolean from Click (defaulting to false). this is essentially what we want:

```
if no_logout is False:
    data.update({"logout_devices": True})
elif no_logout is True:
    data.update({"logout_devices": False})
```

rht: Actually, my `is not None` comparison is moot, because click will assign a default value that is not None anyway?

rht: Safest would be to always specify this "logout_devices" every single time, be it default value or explicitly specificied by the user?

rht: `data.update({"logout_devices": not no_logout}`?

rht: Or rather, just specify it in the initial declaration of `data`.

rht:
When I made this PR, I wasn't aware that Click always specify a default value that is not None. I will make a fresh new PR for the problem.

PR #18, opened 2021-02-06
[BUGFIX] Invert no_logout option when being specified into API

rht:
Like this?

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -72,9 +72,10 @@ class SynapseAdmin:
     def user_password(self, user_id, password, no_logout):
         """ Set the user password, and logs the user out if requested
         """
-        data = {"new_password": password}
-        if no_logout:
-            data.update({"logout_devices": no_logout})
+        data = {
+            "new_password": password,
+            "logout_devices": not no_logout,
+        }
         return self.query("post", f"v1/reset_password/{user_id}", data=data)
 
     def user_details(self, user_id):


JOJ0:
password will always be needed!

synadm/api.py
```diff
@@ -72,9 +72,10 @@ def user_deactivate(self, user_id, gdpr_erase):
...
         """ Set the user password, and logs the user out if requested
         """
-        data = {"new_password": password}
-        if no_logout:
-            data.update({"logout_devices": no_logout})
+        data = {
+            "new_password": password,
+            "logout_devices": not no_logout,
+        }
         return self.query("post", f"v1/reset_password/{user_id}", data=data)
```
JOJ0, line 79: this certainly would work and is correct but it is a general question on how to work with the defaults of the api itself. here we rely on our own application defaults. Up to now I mostly tried to respect the api defaults themselves and just leave out data/things enitrely and let the api take it's default.

but since the beginning of writing this tool I was wondering if this is good practice / best practice or if one should rather rely on app-internal defaults. What if the api changes? (well I guess this is why api providers don't really change it once it's stable, except there is actual bugs that need to be fixed)

I am very interested in what is your opinion on this! :-) For me it's a hard to make decission and I don't know what is right or "righter" ;-)

rht: We don't want users to have to consult the API specs if they want to know the defaults, so it's kind of our responsibility to match the defaults specified in Click with the specs default, and (rarely) change it according to specs if it happens.

Another approach in the implementation is to make sure Click options default to None, and then do the `is not None` check, but in the documentation, we still need to say what the defaults of the API specs are, and update accordingly.

JOJ0: defaults of the cli should be shown in --help. this is lready the case for all (hopefully) commands.

users dont need to read api docs at all 😀

JOJ0: > Another approach in the implementation is to make sure Click options default to None, and then do the `is not None` check, but in the documentation, we still need to say what the defaults of the API specs are, and update accordingly.

I think None this is the default in str options and I use it a lot.

is_flag options are boolean obviously.

I often use the If xxx: pattern to put in our leave out options. As you pointed me to already: All these statements should be corrected to If xxx is not None:

JOJ0: So in this particular case I would suggest to implement like this:
```
if no_logout:
    "logout_devices": False,
```

Then `not no_logout` is difficult to read I think, let's stay pythonic, readability and stuff :-) Also: It is a boolean so "if is not None" does not apply/is not correct.

And I don't know if I got you correctly: In my suggested version here we are relying on that the defaut of the API being "logout_devices: true", so we are definitely relying on it's stability of implementation/default behaviour. Is this good practice in programming design? Is this what you ment with

>  to match the defaults specified in Click with the specs default

Another reason for my suggestion is that I did this kind of "putting together api calls" in probably 80% of cases. .

_If_ changed to your latest suggested version then all the other API implementations should be changed to suit that behaviour to be consistent. Meaning: Not relying on API defaults at all and hardcode every default into synadm.

rht: If you do
```
if no_logout:
    "logout_devices": False
```
you are infusing an implicit assumption in this code that `no_logout` defaults to False in Click. With `"logout_device": not no_logout` <- this line does not need to be aware of what the default value is in Click.

JOJ0: Exactely, because the default in Click _is_ False. I don't see the problem :-))

rht: You are saying the default in Click is always False for all options, not just the one I changed in this PR?

JOJ0: All the code in synadm is relying on Click's default values, and secondly is relying on what defaults the Synapse admin API has, not always all parameters/data needs to be included in the requests. That's just  how it is designed. _If_ I really should I accept changing this method so it does not rely on anything Click does and/or the API does this would be:
* inconsistent
* must be changed in ALL other methods as well
* a hell load of work which I am not willing to do as I need to continue supporting new/all available admin API's, no time for major refactoring of things that just work, sorry!

Or is it what you would like to tell me that you are planning on a PR that refactors the whole SynapseAdmin class to have proper defaults itself that reflect what the (current) defaults of the Synapse admin API are?

JOJ0: I hope you see the bigger picture in this project and sorry for being direct. If you find the time to refactor everything and also test everything, work away. I personally don't have the time to refactor nor to do testing. I need to finish support first. I hope you understand that :-)

rht: Looks like all of the `click.option`s default to False. I grepped for `default=True`. All the result comes from `show_default=True`, none has `default=True`. Since there is no exception/edge case, I think the code is fine the way it is.

JOJ0: > You are saying the default in Click is always False for all options, not just the one I changed in this PR?

The default in Click depends on the data type the option has and the default you code in there. Some examples:

* String options have None as the default or any string you want it to be the default (use default="string argument on the option decorator)
* options with is_flag=True are boolean and have false by default, if you want it to be true, use default=True)

For any further details please consult the click manual pages, they are very detailed and super handy. It is a well-documented and well-maintained project: https://click.palletsprojects.com/en/7.x/

PR #19, opened 2021-02-10
query: Always return resp.json() as long as there's no exception

rht:


JOJ0:
thanks, again: very happy you found this bug! :-)

synadm/api.py
```diff
@@ -39,9 +39,9 @@ def query(self, method, urlpart, params=None, data=None):
                 params=params, json=data
             )
             resp.raise_for_status()
-            if resp.ok:
-                return resp.json()
-            self.log.warning("No valid response from Synapse")
+            if not resp.ok:
```
rht, line 42: @JOJ0 actually the `resp.raise_for_status()` above raises an exception even before `resp.json()` got returned. Should we remove `resp.raise_for_status()`?

rht: #20

JOJ0: Ok I understand but didnt you test this change before submitting the PR??

rht: Sorry, I did test a version of this change, but when I recreated the change in this PR (#19), I somehow missed that there is the `resp.raise_for_status()` line. Let's say it was a lucky accident that I was able to see the failed response json at all, that I somehow put `resp.json()` before `resp.raise_for_status()`.

PR #20, opened 2021-02-11
query: Don't raise during bad request so we know the err msg from server

rht:


JOJ0:
Hi, sorry for the delayed response, being pretty busy at work this week.
Thanks for this PR. This is actually very nice change and improves usability! Played around with it a little and it seems Synapse responses nicely on errors. I did not test each and every API but the one's I tested respond with helpful messages.

Could you do me another favor and also change the wording of the error message and add the status_code too.

Something like:
```
WARNING Synapse returned 400:
errcode: M_UNKNOWN
error: Can only lookup local users
```

synadm/api.py
```diff
@@ -38,7 +38,6 @@ def query(self, method, urlpart, params=None, data=None):
                 url, headers=self.headers, timeout=self.timeout,
                 params=params, json=data
             )
-            resp.raise_for_status()
             if not resp.ok:
                 self.log.warning("No valid response from Synapse")
```
rht, line 42: Did you mean that I should reword this to "WARNING Synapse returned XXX:"? I assume the `resp.json()` output will be pretty-printed by some kind of outputter?

JOJ0: exactley, the resp.json is taken care of already :-)

rht: Updated the warning. Tested, but manually by editing my other PR on `make-admin`.

JOJ0:
thanks so much @rht for another essential fix! appreciated! :-)

PR #23, opened 2021-02-16
Implement the delete group API

aaronraimist:


JOJ0:
Hi thanks a lot for the contribution. I'll try to quickly try it out tonight but from the first looks this can't be wrong and will be merged soon :-) Just being curious: What does the group delete command do? What groups are we talking about? Spaces? Communities? Something else? I think I came across this API while reading docs but couldn't figure out what groups it actually is about. Thanks for shedding some ligth :-)

Also thanks for the readme additions! Great! Very useful!

aaronraimist:
This is deleting what Element calls communities, which were previously called groups. Synapse hasn't really bothered to rename them so they are mostly called groups in Synapse land.

I put the description for `group` as 

> Manage groups (communities)

so hopefully that is clear enough. Otherwise I can add "(communities)" everywhere it says "groups".

Spaces don't exist in Synapse yet.

JOJ0:
Ah alright it's communities. Now that you say it...groups...communities... I think I knew this once haha

Because of naming: Yes I think your suggestion is good. Make it "group (community)" (or plural) everywhere, so no questions and confusions will be left at all. The command name group is actually nice, not only because it reflects what one would find if looking it up in the api docs but also because it's way shorter than community.

Besides that, happy to merge 👍

JOJ0:
you could even add (community) to the link text in the implementation status list

PR #26, opened 2021-02-18
Add new API features into README implementation status since 1.27 release

rht:
These are what I saw from the 1.26..1.27 changelog. So this doesn't include changes that were new in 1.26.


JOJ0:
Hi, very nice! Thanks!

JOJ0:
yeah I think the login-as-a-user-api is also missing. not sure when it went in.

PR #27, opened 2021-02-18
Implement command for user whois

rht:
I was testing a Synapse admin web UI (https://github.com/Awesome-Technologies/synapse-admin
) when I found that I could see all the sessions of a user (including the IP address they used when logging in). This cheat-like feature sounds useful, and so I made this PR.

Not sure if I should add "including IP addresses information" in the docstring.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,7 @@ python3 setup.py install
   * [x] `user deactivate <user id>`
   * [x] `user password <user id>`
   * [x] `user membership <user id>`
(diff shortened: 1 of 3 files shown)


rht:
Definitely a privacy problem: https://github.com/matrix-org/synapse/issues/8473.

JOJ0:
Hey rht that's a neat little feature! Thanks a ton! Merging!

PR #28, opened 2021-02-20
README: Say that `user deactivate` also has GDPR erase option

rht:
In this web UI (https://github.com/Awesome-Technologies/synapse-admin), in one of the screenshot, there is an option to erase user data (for GDPR). I thought synadm doesn't have this feature yet until I grepped the codebase and found that it exists. This feature should be told to the user at the top level documentation.


JOJ0 (approved):
yeah nice idea! GDPR is a concern for a lot of people, often also in corporate context. It should be easy to find! Merging! Thanks!

PR #32, opened 2021-03-01
Implement command for room join

rht:
Was trying to do this from cli, and found that this command hasn't been implemented yet, so I made this PR. Tested on my Matrix instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,8 +167,8 @@ python3 setup.py install
   * [ ] `history purge-status <purge id>`
 * [x] ~~[Purge room API](https://github.com/matrix-org/synapse/blob/master/docs/admin_api/purge_room.md)~~ (DEPRECATED, covered by `room delete`)
 * [ ] [Register API](https://github.com/matrix-org/synapse/blob/master/docs/admin_api/register_api.rst)
-* [ ] [Room membership API](https://github.com/matrix-org/synapse/blob/master/docs/admin_api/room_membership.md)
-  * [ ] `room join`
+* [x] [Room membership API](https://github.com/matrix-org/synapse/blob/master/docs/admin_api/room_membership.md)
+  * [x] `room join`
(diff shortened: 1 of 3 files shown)


JOJ0:
Very nice, looks perfect! Thanks for the submission!

Giving a room alias doesn't work on my server though, but that's surely not the fault of your code!
```
(k_synadm) jojo@jum /tmp/synadm (master) $ synadm room list -n testroom1 | grep alias
 'rooms': [{'canonical_alias': '#testroom1:peek-a-boo.at',
(k_synadm) jojo@jum /tmp/synadm (master) $ synadm room join '#testroom1:peek-a-boo.at' '@jojo:peek-a-boo.at'
{'errcode': 'M_UNKNOWN', 'error': ' was not legal room ID or room alias'}
(k_synadm) jojo@jum /tmp/synadm (master) $
```

Joining with the room id does work flawlessely! Does using canonical_alias work on your end?

rht:
Same here, but as an extra info, when I did `synadm room join 'testing' '@user:domain'` I got "error: testing was not legal room ID or room alias". Notice it is now 'testing' instead of empty string.

rht:
This could be a genuine Synapse bug, but need to gather more info.

JOJ0:
> Same here, but as an extra info, when I did `synadm room join 'testing' '@user:domain'` I got "error: testing was not legal room ID or room alias". Notice it is now 'testing' instead of empty string.

Ah yes now I see it, the error is different, depending on the string, could be that the special character # is confusing Synapse. But you know what: This shouldn't be the concern of this PR. If you want you can make a bugreport on the Synapse repo, but I have to admit that I don't really have time for it and personally would consider it "not worth the hassle".

Your PR looks good to me and I will merge it :-)
Thanks again!
Jojo

rht:
I reported at https://github.com/matrix-org/synapse/issues/9527. I was able to reproduce the problem with a raw curl.

PR #39, opened 2021-07-04
Activating a user without setting a password is a valid user case

andir:
Activating users in an SSO environment is a valid operation even when not providing a password. We shouldn't require a password to be set in those cases.

### DIFF
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -277,10 +277,6 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
     if password_prompt and password:
         click.echo("Use either '-p' or '-P secret', not both.")
         raise SystemExit(1)
-    if deactivation == "activate" and not (password_prompt or password):
-        click.echo("Need to set password when activating a user.")
-        click.echo("Add either '-p' or '-P secret' to your command.")
-        raise SystemExit(1)
     if deactivation == "deactivate" and (password_prompt or password):
         click.echo(
             "Deactivating a user and setting a password doesn't make sense.")


JOJ0:
Hi, thanks a lot for the pull.
I just had a look and realized that I probably misread the last sentence here: https://matrix-org.github.io/synapse/develop/admin_api/user_admin_api.html

```
In order to re-activate an account deactivated must be set to false. If users do not login via single-sign-on, a new password must be provided.
```

So what you are suggesting is actually well documented here. My bad. Thanks a lot for the hint!
What I am thinking now is: How could we make sure that a user that is not in an SSO environment is reactivated but the admin forgets to set a password.

I did not test it yet and also not investigate thoroughly but maybe you know already: On reactivation, does the password that was set prior to deactivation already exist and is valid then? Or would we now have a user that has no password set, and is also allowed to login without one? The latter would be a case I'd like to prevent.

andir:
A new password is required as the password hash is wiped on deactivation: https://github.com/matrix-org/synapse/blob/c1ddbbde4fb948cf740d4c59869157943d3711c6/synapse/handlers/deactivate_account.py#L122

JOJ0:
Please be so kind and spit out a warning to the user when no password is set on re-activation. State that this usually is ok in SSO environments. Or something like that.
The warning should only appear in non-batch mode. See here on how to ask if we are running in batch or not: 
https://github.com/JOJ0/synadm/blob/master/synadm/cli/user.py#L116
https://github.com/JOJ0/synadm/blob/master/synadm/cli/user.py#L456

JOJ0:
hi @andir maybe you didn't get notified because of github notification setting, thus direct mention.
HTH

andir:
I got the notification and I'll get to it as as soon as I can.

On 25 July 2021 17:15:11 CEST, J0J0 T ***@***.***> wrote:
>hi @andir maybe you didn't get notified because of github notification setting, thus direct mention.
>HTH
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/JOJ0/synadm/pull/39#issuecomment-886216052
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

JOJ0:
Hi @andir, can I assist with anything to get this patch merged?

JOJ0:
Hi @andir I hope you don't mind if I merge this and add a warning as suggested? Thanks for the pull.

andir:
On 22:22 04.10.21, J0J0 T wrote:
> Hi @andir I hope you don't mind if I merge this and add a warning as suggested? Thanks for the pull.

Yes, please go ahead. I currently don't have the resources to pick this
back up.

JOJ0:
@andir Never got back to this. Wanted to note: I noticed that on my prod Synapse (latest stable) the API itself spits out an error/a note that a user can't be activate when no password is passed.

On your systems, where SSO is enabled what is the behaviour there?

PR #40, opened 2021-07-05
Fix English grammar in message

neunenak:
Replace "There is <plural noun>" with "There are <plural noun>"
in several messages in the codebase.


JOJ0:
Hi, thanks a lot for the pull.
I guess I was thinking something like "There is more users existing than are shown." but abbreviated it.
But actually now that I think about it: That's actually wrong too and should be "There are more users existing than are shown"
So in any case: Thanks a lot for the correction! Appreciated :-)

Issue #25, opened 2021-02-18
synadm user membership: There should be an option to display room alias if applicable


rht:
As of now whenever I do `synadm user membership @user:host.com`, I got
```
joined_rooms:
- '!oBvIoUsLyNoTaNiD:host.com'
- '!HaHaHa:host.com'
total: 3
```

for yaml output. This can't be made sense of by humans. Better if there is a flag to display aliases (if applicable) instead.

JOJ0:
yes definitely good idea but I need to think about how to realise it. Having a flag that enables this "additional functionality" is essential as one might use the output of these room ID's in consecutive batch outputs....like with using --batch and "jq" filtering.

probably a `room details`command/api call could help gathering the room name!

JOJ0:
@rht have a look, I finally implemented this feature. Thanks for the idea and possibly a review in advance!

PR #41, opened 2021-08-03
Implements room_alias resolution to room_id

MacLemon:
When you need to get the `room_id` from a known `room_alias` you can now use `room resolve <room_alias>`. This is handy when one also uses other tools besides `synadm` that may only use a `room_id` and cannot operate with a `room_alias`.

Example:

```
synadm room resolve "#foo:matrix.org"

-------  ------------------------------
room_id  !jLHcnCpckBMWIfzTOa:matrix.org
servers  ['matrix.org']
-------  ------------------------------
```

### DIFF
--- a/synadm/cli/room.py
+++ b/synadm/cli/room.py
@@ -40,6 +40,16 @@ def join(helper, room_id_or_alias, user_id):
     helper.output(out)
 
 
+@room.command()
+@click.argument("room_alias", type=str)
+@click.pass_obj
+def resolve(helper, room_alias):
+    """ lookup room id from alias
+    """
+    out = helper.matrix_api.resolve(room_alias)
+    helper.output(out)
+
+
 @room.command(name="list")
 @click.pass_obj
 @click.option(
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -32,6 +32,7 @@ import requests
 from http.client import HTTPConnection
 import datetime
 import json
+import urllib.parse
 
 
 class ApiRequest:
@@ -206,6 +207,11 @@ class Matrix(ApiRequest):
             "user": f"{user_id}",
             "initial_device_display_name": "synadm matrix login command"
         })
+
+    def resolve(self, room_alias):
+        """ Allow a user to get the room_id with a given room_alias
+        """
+        return self.query("get", f"client/r0/directory/room/{urllib.parse.quote(room_alias)}")
 
     def raw_request(self, endpoint, method, data, token=None):
         data_dict = {}


JOJ0:
oh wow that's a very nice feature 😍 will look into tomorrow! ☝️

synadm/cli/room.py
```diff
@@ -40,6 +40,16 @@ def join(helper, room_id_or_alias, user_id):
     helper.output(out)
 
 
+@room.command()
+@click.argument("room_alias", type=str)
+@click.pass_obj
+def resolve(helper, room_alias):
+    """ lookup room id from alias
```
JOJ0, line 47: maybe you could add something that states that all servers that joined that room are displayed as well, but in the third line in the docstring so it does only display with `synadm room resolve -h` and not with `synadm room resolve`, keeping it short there is good.

synadm/cli/room.py
```diff
@@ -40,6 +40,16 @@ def join(helper, room_id_or_alias, user_id):
     helper.output(out)
 
 
+@room.command()
+@click.argument("room_alias", type=str)
+@click.pass_obj
+def resolve(helper, room_alias):
```
JOJ0, line 46: I do like the name actually, but am thinking next steps/future commands: You have a suggestion on how the opposite of such a command could be called? As in getting the room alias from a room id. As this was suggested already, well not directly but as an option to room commands directly showing the alias instead of the id: https://github.com/JOJ0/synadm/issues/25. And thus I think a single command translating id to alias would also be nice and a good first step :-)

synadm/cli/room.py
JOJ0: and yeah if you find the time a docstring stating Args: would be nice since only then it will render nicely in the autogenerated sphinx docs. Like here: https://github.com/JOJ0/synadm/pull/41/files#diff-21da8274044fdb3c233e057a58a867c3e934030edd102cbe0a79a61858386d20L194

synadm/cli/room.py
MacLemon: I agree that resolve would fit both directions. Unsure if it's *really* perfect to use it that way though.
`room_alias` to `room_id` will always only return a single `room_id`, while the opposite direction can yield multiple `room_alias`es for a `room_id`.

synadm/cli/room.py
JOJ0: or maybe it could be simply:
```
synadm room getid 'alias'
synadm room getalias 'id'
```
or even
```
synadm room id 'alias'
synadm room alias 'id'
```

synadm/cli/room.py
JOJ0: further brainstorming this morning here ;-) or maybe it could be simply:
```
synadm room getid 'alias'
synadm room getalias 'id'
```
or even
```
synadm room id 'alias'
synadm room alias 'id'
```

JOJ0:
Hi @MacLemon, can I assist with anything to get this feature merged?

JOJ0:
Hi @MacLemon, my suggestion would be to keep this one simple and just add a `-r/--reverse` option that just flips the behaviour: A given room id should be translated into all it's aliases.

Also I'd like to remove the `servers` from the output to make it super simplistic and clear what it does.

Do you mind if I merge the PR and implement the rest like that?

JOJ0:
Hi @MacLemon I implemented as suggested. It's in master. If you find the time to play around with it for a couple of minutes, I'd be happy if you tell me what you think. I hope it's still fitting your initial use-case. Thanks for the pull!

Issue #44, opened 2021-09-27
Force user logout option


ser:
There should be an option to invalidate all access tokens for particular user = force logout.

JOJ0:
> There should be an option to invalidate all access tokens for particular user = force logout.

hi thanks for the idea but is there an API available for that? Did you check the Synapse admin API docs yet? https://matrix-org.github.io/synapse/latest/usage/administration/admin_api/index.html

Synadm is actually basically a cli frontend for what this API's are offering.

JOJ0:
maybe have a look in the user admin section of the docs and see what would suit your usecase. thank

ser:
I suppose it requires two steps:

First get all devices belonging to the user:

The API is:

GET /_synapse/admin/v2/users/<user_id>/devices

And second - delete devices:

The API is:

POST /_synapse/admin/v2/users/<user_id>/delete_devices

{
  "devices": [
    "QBUAZIFURK",
    "AUIECTSRND"
  ],
}

Is it more clear now? :-)

nemobis:
I will make a patch for this.

Use case: mass delete unused devices/sessions for all users in a room, e.g. for performance reasons (encryption cost for each message grows with the number of recipient devices) and to reduce the attack surface.

Acceptance criteria:
* can be chained to a source of user IDs, be it a command line argument or just a previous synadm;
* allow to delete the devices last seen more than N days ago;
* allow to keep at least the M most recent devices.

JOJ0:
> I will make a patch for this.

@nemobis nice! thanks in advance, that's great news! :-)

> Use case: mass delete unused devices/sessions for all users in a room, e.g. for performance reasons (encryption cost for each message grows with the number of recipient devices) and to reduce the attack surface.
> 
> Acceptance criteria:
> 
> * can be chained to a source of user IDs, be it a command line argument or just a previous synadm;

My first idea when I read this feature suggestion was that the delete command should simply accept device id's as returned by https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#list-all-devices, having it accept user id's is probably closer to real use cases. 

Also I was thinking that actually I am seeing 3 commands here: 
- `user device list`
- `user device details`
- `user device delete`

> * allow to delete the devices last seen more than N days ago;
> * allow to keep at least the M most recent devices.

[...]

PR #45, opened 2021-10-05
Add user prune-devices command

nemobis:
Fixes issues #44

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -436,6 +436,19 @@ class SynapseAdmin(ApiRequest):
         """
         return self.query("get", f"v1/whois/{user_id}")
 
+    def user_devices(self, user_id):
+        """ Return information about all devices for a specific user
+        """
+        return self.query("get", f"v2/users/{user_id}/devices")
+
+    def user_devices_delete(self, user_id, devices):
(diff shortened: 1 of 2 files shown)


nemobis:
I think the API returns error 400 when trying to retrieve devices for a non-local user. Not sure what's the best way to handle that.

By the way, initially I used some JSON querying to find filters which would make sense. Something crude like this:

```
for user in `synadm room members <room_id> | jello -rl _.members`; do
  if [[ $user =~ <local_domain> ]]; then
    curl -s --header "Authorization: Bearer $token" "<matrix_domain>/_synapse/admin/v2/users/$user/devices" | jello -l '\
import time
maxms=(round(time.time())-3600*24*90)*1000
ids = []
if _.get("total", 0) > 1:
  for d in _["devices"]:
    seen = d.get("last_seen_ts", None)
    if not seen or seen < maxms:
      ids.append(d.get("device_id", ""))
ids'
```

JOJ0:
hi, @nemobis,

I was wondering what the answer to this question is :-) Thanks!

> The main thing thing that will be most work is the refactoring of some things to api.py as described in one of the 3 inline comments. Do you think you can do that? Not today but in the next couple of days/weeks?

PR #59, opened 2022-03-23
Added support for localpart as <user_id> argument

Ascurius:
Users can now enter the localpart as `<user_id>` argument instead of providing the full MXID

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -304,15 +304,11 @@ class Matrix(ApiRequest):
         """
         if user_id is None:
             return None
-        elif any(x in user_id for x in ["@", ":"]):
-            if re.match(r"\@.+\:.+", user_id):
-                return user_id
-            else:
-                mxid = "@{}:{}".format(re.sub("[@:]", "", user_id),
-                                       self.server_name())
-                return mxid
+        elif re.match(r"@[\w\d\_\.\_\=\-\/]+:[\w\d]+", user_id):
+            return user_id
         else:
-            mxid = "@{}:{}".format(user_id, self.server_name())
+            localpart = re.sub("[@:]", "", user_id)
+            mxid = "@{}:{}".format(localpart, self.server_name())
             return mxid
 
 
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -299,7 +299,7 @@ class Matrix(ApiRequest):
             user_id (string): User ID given by user as command argument.
 
         Returns:
-            string: the fully qualified Matrix User ID (MXID) or None if the 
+            string: the fully qualified Matrix User ID (MXID) or None if the
                 user_id parameter is None.
         """
         if user_id is None:
@@ -308,8 +308,8 @@ class Matrix(ApiRequest):
             if re.match(r"\@.+\:.+", user_id):
                 return user_id
             else:
-                mxid = "@{}:{}".format(re.sub("[@:]", "", user_id), 
-                                    self.server_name())
+                mxid = "@{}:{}".format(re.sub("[@:]", "", user_id),
+                                       self.server_name())
                 return mxid
         else:
             mxid = "@{}:{}".format(user_id, self.server_name())
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -304,7 +304,7 @@ class Matrix(ApiRequest):
         """
         if user_id is None:
             return None
-        elif re.match(r"@[\w\d\_\.\_\=\-\/]+:[\w\d]+", user_id):
+        elif re.match(r"@[-_/=\w\d\.]+:[-_.\w\d]+", user_id):
             return user_id
         else:
             localpart = re.sub("[@:]", "", user_id)
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -291,8 +291,9 @@ class Matrix(ApiRequest):
         return resp['server_name']
 
     def generate_mxid(self, user_id):
-        """ Checks if the given user ID is already a MXID else generating
-            the MXID from the plain user name and the server domain.
+        """ Checks whether the given user ID is an MXID already or else
+        generates it from the passed string and the homeserver name fetched
+        via the server_name method.
 
         Args:
             user_id (string): User ID given by user as command argument.
@@ -304,10 +305,14 @@ class Matrix(ApiRequest):
         if user_id is None:
             return None
         elif any(x in user_id for x in ["@", ":"]):
-            return user_id
+            if re.match(r"\@.+\:.+", user_id):
+                return user_id
+            else:
+                mxid = "@{}:{}".format(re.sub("[@:]", "", user_id), 
+                                    self.server_name())
+                return mxid
         else:
-            mxid = "@{}:{}".format(re.sub("@|:", "", user_id), 
-                                   self.server_name())
+            mxid = "@{}:{}".format(user_id, self.server_name())
             return mxid
 
 
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -304,7 +304,7 @@ class Matrix(ApiRequest):
         """
         if user_id is None:
             return None
-        elif re.match(r"@[-_./=\w\d]+:[-.\w\d]+", user_id):
+        elif re.match(r"@[-./=\w]+:[-.\w]+", user_id):
             return user_id
         else:
             localpart = re.sub("[@:]", "", user_id)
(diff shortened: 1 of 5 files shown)


JOJ0:
Hi @Ascurius, please just merge the tiny fix for ban/unban into master directly and kick it out of this pr. thanks

synadm/api.py
```diff
@@ -289,6 +290,31 @@ def server_name(self):
...
+            user_id (string): User ID given by user as command argument.
+
+        Returns:
+            string: the fully qualified Matrix User ID (MXID) or None if the
+                user_id parameter is None.
+        """
+        if user_id is None:
+            return None
+        elif any(x in user_id for x in ["@", ":"]):
+            if re.match(r"\@.+\:.+", user_id):
```
JOJ0, line 308: I don't think the @ and the : need to be escaped. Would improve readability a little IMHO.

I like that it's even recognized when @localpart or localpart: is the input! Nice!

Also we could narrow down the regex match even more. The problem is I just can't find documentation about which characters are actually allowed in the localpart. Is dash and underscore allowed? I just asked that on #synapse:matrix.org.

Anyway, what shoud definitely work but probabIy would be a little too much allowed still is using the "non whitespace sequence `\S`": `@\S+:\S`

https://docs.python.org/3.7/library/re.html

Ascurius: I agree that the readability is currently not ideal. 

Regarding the rexeg match: The specification of the MXID can be found here: https://spec.matrix.org/v1.2/appendices/#user-identifiers. As stated in the specification, we have to include the following characters: `a-z`, `0-9`, `.`, `_`, `=`, `-`, and `/` as valid input. Therefore I think that the blanket use of `\S` would not be ideal since it also accepts characters like `?` or `+` which are not allowed within the localpart. 
For input validation it would be best to explicitly specify the allowed character, e.g. by using the regex `[\w\s\_\.\_\=\-\/]+` for the localpart. I know that this regex is probably not the best regarding readability but it catches only the allowed characters. Let me know what you think about this approach.

JOJ0: Go for it! Specific is good! 
I think the \s (whitespace) slipped in and shouldnt be there, right?

Ascurius: I have implemented this approach and improved the strcuture of the MXID generation in 406d5792abc9b277c4b3c3741e3db9de01f4fd04. I have already done basic testing, but more in-depth testing is required. Please let me know if you find any issues.

synadm/api.py
```diff
@@ -304,15 +304,11 @@ def generate_mxid(self, user_id):
...
-                return user_id
-            else:
-                mxid = "@{}:{}".format(re.sub("[@:]", "", user_id),
-                                       self.server_name())
-                return mxid
+        elif re.match(r"@[\w\d\_\.\_\=\-\/]+:[\w\d]+", user_id):
+            return user_id
         else:
-            mxid = "@{}:{}".format(user_id, self.server_name())
+            localpart = re.sub("[@:]", "", user_id)
```
JOJ0, line 310: Ah, nice you kicked out the any(x... but kept the substitution of @ and colon here. Nice!

synadm/api.py
```diff
@@ -304,15 +304,11 @@ def generate_mxid(self, user_id):
...
         if user_id is None:
             return None
-        elif any(x in user_id for x in ["@", ":"]):
-            if re.match(r"\@.+\:.+", user_id):
-                return user_id
-            else:
-                mxid = "@{}:{}".format(re.sub("[@:]", "", user_id),
-                                       self.server_name())
-                return mxid
+        elif re.match(r"@[\w\d\_\.\_\=\-\/]+:[\w\d]+", user_id):
```
JOJ0, line 307: In the localpart you have 2 times the underscore. Also some of it don't need to be escaped. I'd suggest this one
`@[-_/=\w\d\.]+`

In the servername part I think we should _at least_ allow everything that's allowed in http urls. So the word-charachters (\w) don't include the dash and the dot. We need to add it. I just realize that I couldn't tell by heart if there's anything else allowed....these days...

Ascurius: Perhaps we can use more information of the server name format as defined in the matrix specification, which can be found here: https://spec.matrix.org/v1.2/appendices/#server-name. 
Let me know if you think we should rely at least in part on the server name specification or if we should only use "normal" domain format for the validation.

synadm/cli/matrix.py
```diff
@@ -56,7 +56,8 @@ def login_cmd(helper, user_id, password):
         else:
             password = click.prompt("Password", hide_input=True)
 
-    login = helper.matrix_api.user_login(user_id, password)
+    mxid = helper.matrix_api.generate_mxid(user_id)
+    login = helper.matrix_api.user_login(mxid, password)
```
JOJ0, line 60: We could enclose the mxid generation inline with the acutally api function call everywhere. I don't have a strong opinion on it, just an idea. I don't think if that's nicer or not, and possible to stay within the 79 characters limit, also it's quite some work to change that everywhere. You decide, it really is just an idea! :-)

```
login = helper.matrix_api.user_login(
    helper.matrix_api.generate_mxid(user_id),
    password
)
```

Ascurius: For now I think that we should leave the structure as it is. The current way where the MXID generation is unraveled from the actual command is a bit more readable in my opinion. As you already mentioned it, for some commands it could be a bit tricky to stay within the 79 characters line limit. On the other hand the format you proposed seems to be more elegant and compact. I will think about your suggestion over the next days and will let you know what my final thoughts are :)

synadm/api.py
Ascurius: Fixed the regex and added domain name validation in 9dfb4652712e3ac67549920ab3e81c304d35d371

synadm/api.py
```diff
@@ -304,7 +304,7 @@ def generate_mxid(self, user_id):
         """
         if user_id is None:
             return None
-        elif re.match(r"@[\w\d\_\.\_\=\-\/]+:[\w\d]+", user_id):
+        elif re.match(r"@[-_/=\w\d\.]+:[-_.\w\d]+", user_id):
```
JOJ0, line 307: localpart is fine now. I think serverpart should be `[-.w\d]+` (without the underscore, because it's not allowed in URL's, but maybe you found out something more about possibly allowed chars. It doesn't hurt too much if we keep the underscore in there I think :-)

I just read the docs again and I think even the dot does not need to be escaped with the backslash because it is inside a set `[abc]`, so either way, we shojuld be consistent how to put in the dot in both the local and the serverpart. Both ways are fine for me and will be understood by regex-aware coders.

Ascurius: > I think serverpart should be [-.\w\d]+ 

Your are right, I will remove the underscore. 

Regarding the `.` in the server part, we should remove the escaping slash. The reason why I am added escaping slashed before every special character, is that I wanted to make absolutely sure that these characteres are only interpreted as plain text. However, I agree that the regex should be consistent, so I suggest that we remove slashes wherever possible.

JOJ0: I just had a closer look again to the the servername matrix spec you posted a couple of days ago I think we should support everything that's stated there but we are on the right track already :-)

But as far as I can see it actually _is_ just what domain names and IP adresses are: 

```
This grammar is based on the standard for internet host names, as specified by [RFC1123, section 2.1](https://tools.ietf.org/html/rfc1123#page-13), with an extension for IPv6 literals.
```

So in short I think this should suffice: `[.-\w\d]`

Ascurius: This is great! I fixed the regex accordingly in ed9b4f6e9115f1b4e09b94e3987706219d2b7335. But I changed the server part to `[-.\w\d]+` since the `-` must be the first character within the set, otherwise an error occurs.

JOJ0: Ah yes, you are right, forgot that already 😀
And just now I also recall that \w includes \d already.
Besides that I think we are ready to merge 🎉

Ascurius: I just figured out that `\w` also includes underscore characters as stated in the documentation (https://docs.python.org/3/howto/regex.html). Therefore I think that we can also remove the explicit underscore.

Ascurius: Just removed the `\d` and the underscore `_` from the regex in [46427c9](https://github.com/JOJ0/synadm/pull/59/commits/46427c9df614a5570ab01e229b454c3618d32c51).

If there is nothing left, we are finally ready to merge.

PR #60, opened 2022-03-27
cli: fall back to str() instead of silence for human output

Xiretza:
When the output consists only of a single value (such as the result of "synadm room resolve"), humanize() returned None, causing nothing at all to be printed.


JOJ0:
Nice catch @Xiretza, you're right falling back to None is a bad idea! I guess that method was written when there was no commands existing that return just one single string or number. Usually it was API responses via `requests`, which always return a list or a dict, at least an empty one.

PR #61, opened 2022-04-14
Fix user_membership in if no aliases are requested

Lykos153:


JOJ0:
Oops that's a nasty one actually! Nice catch @Lykos153, thanks a lot!

PR #63, opened 2022-04-14
Fix NoneType err in server_name method

JOJ0:
- If fetching of key/v2/server doesn't get a response at all (api.query
  returned None already) fail and log immediately instead of trying to look
  into a dict.
- Try to look into the dict in a second step (or).

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -285,7 +285,7 @@ class Matrix(ApiRequest):
                 could not fetch it for any reason.
         """
         resp = self.query("get", "key/v2/server")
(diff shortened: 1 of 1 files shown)


Ascurius:
Should we provide additional information to the user in case the server name could not be fetched, or is it sufficient to only include the information in the logs? Maybe providing a dedicated message to the user? 

Let me know what you think about this. Besides, your changes look good to me.

JOJ0:
log.warning and log.error are always printed on console as well 😀

Issue #68, opened 2022-05-21
Reintroduce the --servername option to the `media` commands 


JOJ0:
...since matrix-docker-ansible-deploy users get failing servername fetches due to the server-server api being used to fetch is not enabled by default.

next steps: auto-fetch servername via different ways:
- .well-known
- dns srv record

JOJ0:
Instead of reintroducing the --servername option above mentioned PR implements a more general approach on retrieval of "the own homeserver name".

PR #69, opened 2022-05-21
Auto-retrieve or use a pre-configured homeserver name

Ascurius:
With regards to #68 this PR aims to change the method of retrieving the server name.

Before the keys/v2/server can be used to retrieve our own servername, it's required to know where the Server-Server aka federation API is located. According to the Matrix spec there is [a definite way to achieving that in the matrix spec
](https://spec.matrix.org/v1.2/server-server-api/#resolving-server-names). It was determined that the admin should be given the freedom to configure the methods (well-known and DNS), rather than implementing an auto-fallback as described in the spec. This should help simplify the implementation and increase performance once set up via synadm config.

As part of this change two new config options `homeserver` and `server_discovery` are utilized.
- `homeserver`: by default set to 'auto-retrieval' or otherwise set to a homeserver name, eg. matrix.DOMAIN already, which overrides auto fetching at all. Eg required if a localhost:port uri is configured for admin api access.
- `server_discovery`: Can either be `dns`, meaning the server name is retrieved using the SRV record, or `well-known`, which indicates that the server name is retrieved using the `.well-known/matrix/server` endpoint. The latter should be the default.

### DIFF
--- a/synadm/cli/__init__.py
+++ b/synadm/cli/__init__.py
@@ -151,6 +151,10 @@ class APIHelper:
             self.config["timeout"], self.requests_debug,
             self.config["server_discovery"], self.config["hostname"]
         )
+        self.misc_request = api.MiscRequest(
+            self.log,
+            self.config["timeout"], self.requests_debug,
+        )
         return True
 
     def write_config(self, config):
--- a/synadm/cli/__init__.py
+++ b/synadm/cli/__init__.py
@@ -226,7 +226,10 @@ class APIHelper:
                     record[0].target, record[0].port
                 )
                 return self.matrix_api.server_name_keys_api(federation_uri)
-            return None
+        else:
+            self.log.error("Unknown server_discovery mode. "
+                           "Launch synadm config!")
+        return None
 
     def generate_mxid(self, user_id):
         """ Checks whether the given user ID is an MXID already or else
--- a/synadm/cli/__init__.py
+++ b/synadm/cli/__init__.py
@@ -182,6 +182,25 @@ class APIHelper:
         """
         click.echo(self.formatter(data))
 
+    def retrieve_homeserver_name(self, uri):
+        """Try to retrieve the homeserver name via .well-known and a
+        Server-Server (Federation) API.
+
+        FIXME: Fall back to a statically configured name in config.yaml or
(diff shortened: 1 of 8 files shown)


JOJ0:
@Ascurius do you mind if I edit your inital description of the feature to fit what we discussed internally and partly is implemented already?

Ascurius:
Of course, go ahead, feel free to update the description

synadm/cli/__init__.py
```diff
@@ -175,6 +183,48 @@ def output(self, data):
...
+        """
+        if self.config["homeserver"] != "auto-retrieval":
+            return  self.config["homeserver"]
+
+        if self.config["server_discovery"] == "well-known":
+            federation_uri = self.misc_request.federation_uri_well_known(uri)
+            if not federation_uri:
+                return None
+            return self.matrix_api.server_name_keys_api(federation_uri)
+        elif self.config["server_discovery"] == "dns":
```
JOJ0, line 209: @Ascurius please review and test. I tried to add some error handling to the DNS SRV resolving. Also tried to make use of the possibilities (attributes target, port) of the dns.resolver object in the else branch of the try/except clause. No extra sanity check of whether a record 0 is there. Do you think that is sufficient? Please do some testing with bollocks values for hostname and simulated network errors.

Ascurius: Sry for my late response. I have reviewed and tested your changes. As far as I can see, everything works as intended, I could not find any bugs 👍

JOJ0: Ah one thing to look closely: I might have forgotten to document newly implemented args in docstrings (or adapt to changed behaviour) Please doublecheck! Thanks a ton!!

JOJ0:
FIXME: We need some kind of "nice error message" when http://localhost:8008 is used when fetching server-name via the new method and thus fetching fails, Something like: Use synadm config to manually set your homeserver name, auto-retrieval can't be used with your setup"

JOJ0:
@Ascurius added some magic to make auto-retrieve work with localhost :-) I think I'll merge this now, ask around for some more testers but think it's pretty close to ready to release, maybe some tiny polishing required.

PR #72, opened 2022-08-15
Fix --server-discovery help and a

JOJ0:
wrong variable name in the config subcommand.


JOJ0:
Hi @Ascurius, found a wrong variable name while trying to describe what the --server-discovery optiona is all about. Thanks for a review in advance!

Ascurius (approved):
Hi @JOJ0, thank you for the rephrasing of the help text. It is now much clearer and understandable what the option does. Thus I would say we are ready to merge this change

PR #73, opened 2022-08-16
Redact token in configurator

JOJ0:


JOJ0:
Hi @Ascurius, the token should never be shown in the configurator but it should be obvious whether a token IS configured or IS NOT.
Does that make sense? Please play around with it a little and tell me if you feel it's superclear from a user perspective.

As a sidenote: What's still not possible is: Setting the token to "empty string", but currently I don't consider that important. It would be important if finally I would start implementing the possibility to
* _only_ set user in the configurator
* no token
* and have the user be prompted for a password for each synadm call (The supersecure option if an occasional synadm user prefers that instead of having to save the token on a system)

But that sidenote is just dumping my "future thoughts" here and shouldn't slow down this PR ;-) Your thoughts on it are welcome certainly! Thanks!

Ascurius:
Hi @JOJ0, I have tested your changes and on my machine, the token was correctly redacted both if it was seit through CLI or by using the configurator. But I think that in the case of an empty token, the configurator should not just show ``[]`` but should include a message for the user similar to ```"NOT SET"```. This would make it absolutely clear that the token was not set yet. Therefore I have commited a correspondig tiny change. Let me know what you think about this proposal.

Regarding your future thoughts, I am convinced of the benefits that your proposals would have for the users. Especially the prompting for a password to authenticate an API call. I will get in touch with you regarding further details.

PR #74, opened 2022-09-12
Add server notice support

a-0-dev:
Implemented features (for now):
- Send to single matrix IDs or to all mxids matching a regular expression
- Uses pagination when sending to multiple users (batches of 100 by default, tunable by CLI option)
- Specify notice content in command arguments or in files
- Supports both unformatted and unformatted+formatted message content
- Asks for confirmation (including message content and sample mxids the notice will be sent to) by default before sending - can be disabled by flag

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -1171,9 +1171,10 @@ class SynapseAdmin(ApiRequest):
                     if not re.search(receivers, user["name"]) is None:
                         data["user_id"] = user["name"]
                         outputs.append(
-                            self.query("post",
-                                       f"v1/send_server_notice", data=data)
-                                      )
+                            self.query(
+                                "post", f"v1/send_server_notice", data=data
+                            )
+                        )
 
                 if "next_token" not in response:
                     return outputs
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -1182,5 +1182,5 @@ class SynapseAdmin(ApiRequest):
                                           100, True, False, "", "")
         # Only a single user ID was supplied as receiver
         else:
-            data["user_id"] = receivers
+            data["user_id"] = generate_mxid(receivers)
             return [self.query("post", f"v1/send_server_notice", data=data)]
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -1168,7 +1168,7 @@ class SynapseAdmin(ApiRequest):
                 return
             while True:
                 for user in response["users"]:
-                    if not re.search(receivers, user["name"]) is None:
+                    if re.match(receivers, user["name"]):
                         data["user_id"] = user["name"]
                         outputs.append(
                             self.query(
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -34,6 +34,8 @@ import datetime
 import json
 import urllib.parse
 import time
+import re
+import click
 
 
 class ApiRequest:
@@ -1139,3 +1141,58 @@ class SynapseAdmin(ApiRequest):
         else:
             method = "post"
         return self.query(method, f"v1/users/{user_id}/shadow_ban")
+
+    def notice_send(self, receivers, content_plain, content_html, batch, paginate):
+        """ Send server notices.
+
+        Args:
+            receivers: Target(s) of the notice. Either localpart or regular 
+                expression matching localparts.
+            content_plain: Unformatted text of the notice.
+            content_html: HTML-formatted text of the notice.
+        """
+        data = {
+                "user_id": "",
+                "content": {
+                        "msgtype": "m.text",
+                        "body": content_plain,
+                        "format": "org.matrix.custom.html",
+                        "formatted_body": content_html
+                }
+        }
+        def ask_confirmation(users):
+            if batch:
+                return True
+            prompt = "Recipients (list may be incomplete):\n"
+            ctr = 0
+            for user in users:
+                if not re.search(receivers, user) is None:
+                    prompt = prompt + " - " + user + "\n"
+                    ctr = ctr + 1
+                    if ctr == 10:
+                        prompt = prompt + " - ..."
+                        break
+            prompt = prompt + "\nUnformatted message:\n---\n" + content_plain + "\n---\nFormatted message:\n---\n" + content_html + "\n---\nSend now?"
+            return click.confirm(prompt)
+
+        # Only a single user ID was supplied as receiver
+        if receivers[:1] == '@':
+            if ask_confirmation([receivers]):
+                data["user_id"] = receivers
+                return [self.query("post", f"v1/send_server_notice", data=data)]
+        # A regular expression was supplied to match receivers.
+        elif receivers[:1] == '^':
+            outputs = []
+            response = self.user_list(0, paginate, True, False, "", "")
+            if ask_confirmation(map(lambda u: u["name"], response["users"])):
+                while True:
+                    for user in response["users"]:
+                        if not re.search(receivers, user["name"]) is None:
+                            data["user_id"] = user["name"]
+                            outputs.append(self.query("post", f"v1/send_server_notice", data=data))
+
+                    if not "next_token" in response:
+                        return outputs
+                    response = self.user_list(response["next_token"], 100, True, False, "", "")
+        else:
+            print("The recipients you specified were neither a matrix ID (starting with an '@'), nor a regular expression (starting with '^').")
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -1144,19 +1144,19 @@ class SynapseAdmin(ApiRequest):
         """ Send server notices.
 
         Args:
-            receivers: Target(s) of the notice. Either localpart or regular 
+            receivers: Target(s) of the notice. Either localpart or regular
                 expression matching localparts.
             content_plain: Unformatted text of the notice.
             content_html: HTML-formatted text of the notice.
         """
         data = {
-                "user_id": "",
-                "content": {
-                        "msgtype": "m.text",
-                        "body": content_plain,
-                        "format": "org.matrix.custom.html",
-                        "formatted_body": content_html
-                }
+            "user_id": "",
+            "content": {
+                "msgtype": "m.text",
+                "body": content_plain,
+                "format": "org.matrix.custom.html",
+                "formatted_body": content_html
+            }
         }
 
         # A regular expression was supplied to match receivers.
@@ -1167,13 +1167,15 @@ class SynapseAdmin(ApiRequest):
                 for user in response["users"]:
                     if not re.search(receivers, user["name"]) is None:
                         data["user_id"] = user["name"]
-                        outputs.append(self.query("post", 
-                            f"v1/send_server_notice", data=data))
+                        outputs.append(
+                            self.query("post",
+                                       f"v1/send_server_notice", data=data)
+                                      )
 
-                if not "next_token" in response:
+                if "next_token" not in response:
                     return outputs
-                response = self.user_list(response["next_token"], 
-                    100, True, False, "", "")
+                response = self.user_list(response["next_token"],
+                                          100, True, False, "", "")
         # Only a single user ID was supplied as receiver
         else:
             data["user_id"] = receivers
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -36,6 +36,7 @@ import urllib.parse
 import time
 import re
 
+
 class ApiRequest:
     """Basic API request handling and helper utilities
 
@@ -1140,7 +1141,7 @@ class SynapseAdmin(ApiRequest):
             method = "post"
         return self.query(method, f"v1/users/{user_id}/shadow_ban")
 
-    def notice_send(self, receivers, content_plain, content_html, paginate):
+    def notice_send(self, receivers, content_plain, content_html, paginate, to_regex):
         """ Send server notices.
 
         Args:
@@ -1160,9 +1161,11 @@ class SynapseAdmin(ApiRequest):
         }
 
         # A regular expression was supplied to match receivers.
-        if receivers[:1] == '^':
+        if to_regex:
             outputs = []
             response = self.user_list(0, paginate, True, False, "", "")
+            if "users" not in response:
+                return
             while True:
                 for user in response["users"]:
                     if not re.search(receivers, user["name"]) is None:
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -35,8 +35,6 @@ import json
 import urllib.parse
 import time
 import re
-import click
-
 
 class ApiRequest:
     """Basic API request handling and helper utilities
@@ -1142,7 +1140,7 @@ class SynapseAdmin(ApiRequest):
             method = "post"
         return self.query(method, f"v1/users/{user_id}/shadow_ban")
 
-    def notice_send(self, receivers, content_plain, content_html, batch, paginate):
+    def notice_send(self, receivers, content_plain, content_html, paginate):
         """ Send server notices.
 
         Args:
@@ -1160,39 +1158,23 @@ class SynapseAdmin(ApiRequest):
                         "formatted_body": content_html
                 }
         }
-        def ask_confirmation(users):
-            if batch:
-                return True
-            prompt = "Recipients (list may be incomplete):\n"
-            ctr = 0
-            for user in users:
-                if not re.search(receivers, user) is None:
-                    prompt = prompt + " - " + user + "\n"
-                    ctr = ctr + 1
-                    if ctr == 10:
-                        prompt = prompt + " - ..."
-                        break
-            prompt = prompt + "\nUnformatted message:\n---\n" + content_plain + "\n---\nFormatted message:\n---\n" + content_html + "\n---\nSend now?"
-            return click.confirm(prompt)
 
-        # Only a single user ID was supplied as receiver
-        if receivers[:1] == '@':
-            if ask_confirmation([receivers]):
-                data["user_id"] = receivers
-                return [self.query("post", f"v1/send_server_notice", data=data)]
         # A regular expression was supplied to match receivers.
-        elif receivers[:1] == '^':
+        if receivers[:1] == '^':
             outputs = []
             response = self.user_list(0, paginate, True, False, "", "")
-            if ask_confirmation(map(lambda u: u["name"], response["users"])):
-                while True:
-                    for user in response["users"]:
-                        if not re.search(receivers, user["name"]) is None:
-                            data["user_id"] = user["name"]
-                            outputs.append(self.query("post", f"v1/send_server_notice", data=data))
+            while True:
+                for user in response["users"]:
+                    if not re.search(receivers, user["name"]) is None:
+                        data["user_id"] = user["name"]
+                        outputs.append(self.query("post", 
+                            f"v1/send_server_notice", data=data))
 
-                    if not "next_token" in response:
-                        return outputs
-                    response = self.user_list(response["next_token"], 100, True, False, "", "")
+                if not "next_token" in response:
+                    return outputs
+                response = self.user_list(response["next_token"], 
+                    100, True, False, "", "")
+        # Only a single user ID was supplied as receiver
         else:
-            print("The recipients you specified were neither a matrix ID (starting with an '@'), nor a regular expression (starting with '^').")
+            data["user_id"] = receivers
+            return [self.query("post", f"v1/send_server_notice", data=data)]
--- a/synadm/cli/notice.py
+++ b/synadm/cli/notice.py
@@ -20,36 +20,40 @@
 
 import re
 import click
-from click_option_group import optgroup, MutuallyExclusiveOptionGroup
-from click_option_group import RequiredAnyOptionGroup
 
 from synadm import cli
+
 
 @cli.root.group()
 def notice():
     """ Send server notices to local users
     """
 
+
 @notice.command(name="send")
-@click.option("--from-file", "-f", default=False, show_default=True
+@click.option("--from-file", "-f", default=False, show_default=True,
     is_flag=True, help="""Interpret arguments as file paths instead of notice
     content and read content from those files""")
 @click.option("--paginate", type=int, default=100, show_default=True,
     help="Sets how many users will be retrieved from the server at once")
 @click.option("--to-regex", "-r", default=False, show_default=True,
     is_flag=True, help="Interpret TO as regular expression")
+@click.option("--match-list-length", type=int, default=10, show_default=True,
+    help="""Length of the displayed list of matched receivers. Does not impact
+    sending behavior""")
 @click.argument("to", type=str, default=None)
 @click.argument("plain", type=str, default=None)
 @click.argument("formatted", type=str, default=None, required=False)
 @click.pass_obj
-def notice_send_cmd(helper, from_file, paginate, to_regex, to, plain, formatted):
+def notice_send_cmd(helper, from_file, paginate, to_regex, match_list_length,
+                    to, plain, formatted):
     """Send server notices to local users.
 
     TO - localpart or full matrix ID of the notice receiver. If --to-regex is
         set this will be interpreted as regular expression.
-    
+
     PLAIN - plain text content of the notice
-    
+
     FORMATTED - Formatted content of the notice. If not set, PLAIN will be
         used.
     """
@@ -67,33 +71,50 @@ def notice_send_cmd(helper, from_file, paginate, to_regex, to, plain, formatted)
             formatted_content = plain
         else:
             formatted_content = formatted
-    
-    def confirm_prompt(users):
-            if helper.batch:
-                return True
-            prompt = "Recipients (list may be incomplete):\n"
+
+    def confirm_prompt():
+        if helper.batch:
+            return True
+        prompt = "Recipients:\n"
+        if not to_regex:
+            prompt = prompt + " - " + to + "\n"
+        else:
+            # Build and print a list of receivers matching the regex
             ctr = 0
-            for user in users:
-                if not re.search(to, user) is None:
-                    prompt = prompt + " - " + user + "\n"
-                    ctr = ctr + 1
-                    if ctr >= 10:
-                        prompt = prompt + " - ..."
-                        break
-            prompt = prompt + "\Plaintext message:\n---\n" + plain_content\
-                + "\n---\nFormatted message:\n---\n" + formatted_content\
-                + "\n---\nSend now?"
-            return click.confirm(prompt)
-    
+            next_token = 0
+            # Outer loop: If fetching >1 pages of users is required
+            while ctr < match_list_length:
+                batch = helper.api.user_list(
+                    next_token, paginate, True, False, "", "")
+                if "users" not in batch:
+                    break
+                batch_mxids = [user['name'] for user in batch["users"]]
+                # Match every matrix ID of this batch
+                for mxid in batch_mxids:
+                    if not re.search(to, mxid) is None:
+                        if ctr < match_list_length:
+                            prompt = prompt + " - " + mxid + "\n"
+                            ctr = ctr + 1
+                        else:
+                            prompt = prompt + " - ...\n"
+                            break
+                if "next_token" not in batch:
+                    break
+                next_token = batch["next_token"]
+        prompt = prompt + "\nPlaintext message:\n---\n" + plain_content\
+            + "\n---\nFormatted message:\n---\n" + formatted_content\
+            + "\n---\nSend now?"
+        return click.confirm(prompt)
+
     if to_regex:
-        first_batch = helper.api.user_list(0, 100, True, False, "", "")
-        if "users" not in first_batch:
+        if "users" not in helper.api.user_list(0, 100, True, False, "", ""):
             return
-        if not confirm_prompt([user['name'] for user in first_batch["users"]]):
+        if not confirm_prompt():
             return
     else:
         to = helper.generate_mxid(to)
-        if not confirm_prompt([to]):
+        if not confirm_prompt():
             return
-    
-    helper.api.notice_send(to, plain_content, formatted_content, paginate, to_regex)
+
+    helper.api.notice_send(to, plain_content, formatted_content, paginate,
+                           to_regex)
--- a/synadm/cli/notice.py
+++ b/synadm/cli/notice.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # synadm
-# Copyright (C) 2020-2022 Philip (a-0)
+# Copyright (C) 2022 Philip (a-0)
 #
 # synadm is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -18,6 +18,7 @@
 """server notice-related CLI commands
 """
 
+import re
 import click
 from click_option_group import optgroup, MutuallyExclusiveOptionGroup
 from click_option_group import RequiredAnyOptionGroup
@@ -31,32 +32,61 @@ def notice():
 
 @notice.command(name="send")
 @click.option("--from-file/--from-argument", "-f/-a", default=False,
-    help="Interpret arguments as file paths instead of notice content and read content from those files")
-@click.option("--batch", "--non-interactive", is_flag=True, default=False,
-    help="Skip the confirmation step before sending notices")
+    help="""Interpret arguments as file paths instead of notice content and 
+    read content from those files""")
 @click.option("--paginate", type=int, default=100,
     help="Sets how many users will be retrieved from the server at once")
 @click.argument("to", type=str, default=None)
 @click.argument("plain", type=str, default=None)
 @click.argument("formatted", type=str, default=None, required=False)
 @click.pass_obj
-def notice_send_cmd(helper, from_file, batch, paginate, to, plain, formatted):
+def notice_send_cmd(helper, from_file, paginate, to, plain, formatted):
     """Send server notices to local users.
 
-    TO - either a matrix ID (e.g. '@abc:example.com') or a regular expression as used in python (e.g. '^.*' to send to all users)
+    TO - either a matrix ID (e.g. '@abc:example.com') or a regular expression 
+        as used in python (e.g. '^.*' to send to all users)
     
     PLAIN - plain text content of the notice
     
-    FORMATTED - (HTML-) formatted content of the notice. If not set, PLAIN will be used.
+    FORMATTED - (HTML-) formatted content of the notice. If not set, PLAIN 
+        will be used.
     """
     if from_file:
-        plain_content = open(plain, "r").read()
+        with open(plain, "r") as plain_file:
+            plain_content = plain_file.read()
         if formatted:
-            formatted_content = open(formatted, "r").read()
+            with open(formatted, "r") as formatted_file:
+                formatted_content = formatted_file.read()
         else:
             formatted_content = plain_content
     else:
         plain_content = plain
         formatted_content = formatted
     
-    helper.api.notice_send(to, plain_content, formatted_content, batch, paginate)
+    def confirm_prompt(users):
+            if helper.batch:
+                return True
+            prompt = "Recipients (list may be incomplete):\n"
+            ctr = 0
+            for user in users:
+                if not re.search(to, user) is None:
+                    prompt = prompt + " - " + user + "\n"
+                    ctr = ctr + 1
+                    if ctr >= 10:
+                        prompt = prompt + " - ..."
+                        break
+            prompt = prompt + "\nUnformatted message:\n---\n" + plain_content\
+                + "\n---\nFormatted message:\n---\n" + formatted_content\
+                + "\n---\nSend now?"
+            return click.confirm(prompt)
+    
+    if to[:1] == '^':
+        first_batch = helper.api.user_list(0, paginate, True, False, "", "")
+        if not confirm_prompt(map(lambda u: u["name"], first_batch["users"])):
+            return
+    else:
+        to = helper.generate_mxid(to)
+        if not confirm_prompt([to]):
+            return
+    
+    helper.api.notice_send(to, plain_content, formatted_content, paginate)
--- a/synadm/cli/notice.py
+++ b/synadm/cli/notice.py
@@ -64,21 +64,6 @@ def notice_send_cmd(helper, from_file, paginate, to_regex, match_list_length,
     FORMATTED - Formatted content of the notice. If not set, PLAIN will be
         used.
     """
-    if from_file:
-        with open(plain, "r") as plain_file:
-            plain_content = plain_file.read()
-        if formatted:
-            with open(formatted, "r") as formatted_file:
-                formatted_content = formatted_file.read()
-        else:
-            formatted_content = plain_content
-    else:
-        plain_content = plain
-        if formatted is None:
-            formatted_content = plain
-        else:
-            formatted_content = formatted
-
     def confirm_prompt():
         if helper.batch:
             return True
@@ -112,6 +97,21 @@ def notice_send_cmd(helper, from_file, paginate, to_regex, match_list_length,
             + "\n---\nFormatted message:\n---\n" + formatted_content\
             + "\n---\nSend now?"
         return click.confirm(prompt)
+
+    if from_file:
+        with open(plain, "r") as plain_file:
+            plain_content = plain_file.read()
+        if formatted:
+            with open(formatted, "r") as formatted_file:
+                formatted_content = formatted_file.read()
+        else:
+            formatted_content = plain_content
+    else:
+        plain_content = plain
+        if formatted is None:
+            formatted_content = plain
+        else:
+            formatted_content = formatted
 
     if to_regex:
         if "users" not in helper.api.user_list(0, 100, True, False, "", ""):
--- a/synadm/cli/notice.py
+++ b/synadm/cli/notice.py
@@ -36,15 +36,17 @@ def notice():
     read content from those files""")
 @click.option("--paginate", type=int, default=100,
     help="Sets how many users will be retrieved from the server at once")
+@click.option("--to-regex", "-r", default=False, show_default=True, 
+    is_flag=True, help="Interpret TO as regular expression")
 @click.argument("to", type=str, default=None)
 @click.argument("plain", type=str, default=None)
 @click.argument("formatted", type=str, default=None, required=False)
 @click.pass_obj
-def notice_send_cmd(helper, from_file, paginate, to, plain, formatted):
+def notice_send_cmd(helper, from_file, paginate, to_regex, to, plain, formatted):
     """Send server notices to local users.
 
-    TO - either a matrix ID (e.g. '@abc:example.com') or a regular expression 
-        as used in python (e.g. '^.*' to send to all users)
+    TO - localpart or full matrix ID of the notice receiver. If --to-regex is
+        set this will be interpreted as regular expression.
     
     PLAIN - plain text content of the notice
     
@@ -61,7 +63,10 @@ def notice_send_cmd(helper, from_file, paginate, to, plain, formatted):
             formatted_content = plain_content
     else:
         plain_content = plain
-        formatted_content = formatted
+        if formatted is None:
+            formatted_content = plain
+        else:
+            formatted_content = formatted
     
     def confirm_prompt(users):
             if helper.batch:
@@ -80,8 +85,10 @@ def notice_send_cmd(helper, from_file, paginate, to, plain, formatted):
                 + "\n---\nSend now?"
             return click.confirm(prompt)
     
-    if to[:1] == '^':
+    if to_regex:
         first_batch = helper.api.user_list(0, paginate, True, False, "", "")
+        if "users" not in first_batch:
+            return
         if not confirm_prompt([user['name'] for user in first_batch["users"]]):
             return
     else:
@@ -89,4 +96,4 @@ def notice_send_cmd(helper, from_file, paginate, to, plain, formatted):
         if not confirm_prompt([to]):
             return
     
-    helper.api.notice_send(to, plain_content, formatted_content, paginate)
+    helper.api.notice_send(to, plain_content, formatted_content, paginate, to_regex)
--- a/synadm/cli/notice.py
+++ b/synadm/cli/notice.py
@@ -31,12 +31,12 @@ def notice():
     """
 
 @notice.command(name="send")
-@click.option("--from-file/--from-argument", "-f/-a", default=False,
-    help="""Interpret arguments as file paths instead of notice content and 
-    read content from those files""")
-@click.option("--paginate", type=int, default=100,
+@click.option("--from-file", "-f", default=False, show_default=True
+    is_flag=True, help="""Interpret arguments as file paths instead of notice
+    content and read content from those files""")
+@click.option("--paginate", type=int, default=100, show_default=True,
     help="Sets how many users will be retrieved from the server at once")
-@click.option("--to-regex", "-r", default=False, show_default=True, 
+@click.option("--to-regex", "-r", default=False, show_default=True,
     is_flag=True, help="Interpret TO as regular expression")
 @click.argument("to", type=str, default=None)
 @click.argument("plain", type=str, default=None)
@@ -50,8 +50,8 @@ def notice_send_cmd(helper, from_file, paginate, to_regex, to, plain, formatted)
     
     PLAIN - plain text content of the notice
     
-    FORMATTED - (HTML-) formatted content of the notice. If not set, PLAIN 
-        will be used.
+    FORMATTED - Formatted content of the notice. If not set, PLAIN will be
+        used.
     """
     if from_file:
         with open(plain, "r") as plain_file:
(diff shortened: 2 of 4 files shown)


synadm/api.py
```diff
@@ -1139,3 +1141,58 @@ def user_shadow_ban(self, user_id, unban):
...
+        data = {
+                "user_id": "",
+                "content": {
+                        "msgtype": "m.text",
+                        "body": content_plain,
+                        "format": "org.matrix.custom.html",
+                        "formatted_body": content_html
+                }
+        }
+        def ask_confirmation(users):
```
JOJ0, line 1163: in api.py there shouldnt be any frontend code, eg user interaction. Move that into cli.notice module

synadm/api.py
```diff
@@ -34,6 +34,8 @@
 import json
 import urllib.parse
 import time
+import re
+import click
```
JOJ0, line 38: in api.py there shouldnt be any frontend code, thus remove click import

synadm/api.py
```diff
@@ -1139,3 +1141,58 @@ def user_shadow_ban(self, user_id, unban):
...
+                if not re.search(receivers, user) is None:
+                    prompt = prompt + " - " + user + "\n"
+                    ctr = ctr + 1
+                    if ctr == 10:
+                        prompt = prompt + " - ..."
+                        break
+            prompt = prompt + "\nUnformatted message:\n---\n" + content_plain + "\n---\nFormatted message:\n---\n" + content_html + "\n---\nSend now?"
+            return click.confirm(prompt)
+
+        # Only a single user ID was supplied as receiver
```
JOJ0, line 1178: in api.py you will find some helper method around usernames. synadm now supports automagically determining if a username is only localpart or is a full mxid. use the eisting helper please.

synadm/cli/notice.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
+# synadm
+# Copyright (C) 2020-2022 Philip (a-0)
```
JOJ0, line 3: 2022 only

synadm/cli/notice.py
```diff
@@ -0,0 +1,62 @@
...
+
+@cli.root.group()
+def notice():
+    """ Send server notices to local users
+    """
+
+@notice.command(name="send")
+@click.option("--from-file/--from-argument", "-f/-a", default=False,
+    help="Interpret arguments as file paths instead of notice content and read content from those files")
+@click.option("--batch", "--non-interactive", is_flag=True, default=False,
```
JOJ0, line 35: Remove this option and use the global option --batch. you can access its value from helper.batch attribute.

JOJ0 (changes requested):
Submitting a first
review. In general have a look on how things are done in the user modify command. It will help understand my suggestions on what should live in the cli module (frontend) and what in api.py (backend). Eg batch and confirmation stuff. Make use of as much methods and attributes from helper. Check its possibilities in cli.__init__

synadm/api.py
Ascurius: Please also make sure that you do not exceed the maximum line length of 79 characters.

synadm/cli/notice.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    if from_file:
+        plain_content = open(plain, "r").read()
+        if formatted:
+            formatted_content = open(formatted, "r").read()
+        else:
+            formatted_content = plain_content
+    else:
+        plain_content = plain
+        formatted_content = formatted
```
Ascurius, line 60: If you open files to read their content, it is generally considered best practice that you explicitly close these files after use (e.g. `file.close()`). Alternatively, you can use a context manager that closes the file itself after use. See [here](https://book.pythontips.com/en/latest/context_managers.html) if you need a resource.

synadm/api.py
```diff
@@ -1139,3 +1139,42 @@ def user_shadow_ban(self, user_id, unban):
...
+        """
+        data = {
+                "user_id": "",
+                "content": {
+                        "msgtype": "m.text",
+                        "body": content_plain,
+                        "format": "org.matrix.custom.html",
+                        "formatted_body": content_html
+                }
+        }
```
JOJ0, line 1160: Use 4 spaces for the initial indentation and then multiples of 4 for the deeper levels of the dict.

synadm/api.py
```diff
@@ -35,8 +35,6 @@
 import urllib.parse
 import time
 import re
-import click
-
```
JOJ0, line 37: Leave the second empty line.

synadm/cli/notice.py
```diff
@@ -31,32 +32,61 @@ def notice():
...
+                        prompt = prompt + " - ..."
+                        break
+            prompt = prompt + "\nUnformatted message:\n---\n" + plain_content\
+                + "\n---\nFormatted message:\n---\n" + formatted_content\
+                + "\n---\nSend now?"
+            return click.confirm(prompt)
+    
+    if to[:1] == '^':
+        first_batch = helper.api.user_list(0, paginate, True, False, "", "")
+        if not confirm_prompt(map(lambda u: u["name"], first_batch["users"])):
```
JOJ0, line 85: I wonder if this could be rewritten without the use of lambda? Even if it would be slightly longer code I'd prefer not useing lambda and map() if not absolutely necessary.

JOJ0 (commented):
I'm so sorry, still on vacation and it's hard for me to find time to review and even test. Just some things I came accross. Will do a proper review when I'm at home again next week.
Please use code formatting standards mostly found in the pep8 standard. I have flake8 running with my code editor but you could `pip install flake8` and then run from terminal for each file: `flake8 synadm/api.py`.  Thanks and sorry for being picky again! We try to keep code style similar as good as possible, because we think it eases contribution.

synadm/api.py
a-0-dev: There is no second empty line any more... the click import has already been removed (?) I double checked, there's exactly one empty line between `import re` and `class ApiRequest:`

JOJ0: Please _leave_ the second blank line present. We'd like to have 2 blank lines between imports and code (like we have it between functions and methods). That's also what pep8 suggests.

synadm/cli/notice.py
JOJ0: Wonderful! Thanks for changing this to a list comprehension. That's much more pythonic and fits the style we have in other places in the project. 👍

JOJ0 (approved):

synadm/cli/notice.py
```diff
@@ -0,0 +1,92 @@
...
+        if formatted:
+            with open(formatted, "r") as formatted_file:
+                formatted_content = formatted_file.read()
+        else:
+            formatted_content = plain_content
+    else:
+        plain_content = plain
+        formatted_content = formatted
+    
+    def confirm_prompt(users):
```
JOJ0, line 66: I'm getting this on my rather old Synapse installation. It might happen because the notice admin API is not present and returning a 404. We could try to catch non-supported Synapse versions early and throw a note to the user. 

```
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send "blabla" @testuser1:peek-a-boo.at
Traceback (most recent call last):
  File "/home/jojo/.venvs/synadm/bin/synadm", line 11, in <module>
    load_entry_point('synadm', 'console_scripts', 'synadm')()
  File "/home/jojo/.venvs/synadm/lib/python3.6/site-packages/click-7.1.2-py3.6.egg/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/jojo/.venvs/synadm/lib/python3.6/site-packages/click-7.1.2-py3.6.egg/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/jojo/.venvs/synadm/lib/python3.6/site-packages/click-7.1.2-py3.6.egg/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jojo/.venvs/synadm/lib/python3.6/site-packages/click-7.1.2-py3.6.egg/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jojo/.venvs/synadm/lib/python3.6/site-packages/click-7.1.2-py3.6.egg/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/jojo/.venvs/synadm/lib/python3.6/site-packages/click-7.1.2-py3.6.egg/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/jojo/.venvs/synadm/lib/python3.6/site-packages/click-7.1.2-py3.6.egg/click/decorators.py", line 33, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/home/jojo/git/a-0_synadm/synadm/cli/notice.py", line 89, in notice_send_cmd
    if not confirm_prompt([to]):
  File "/home/jojo/git/a-0_synadm/synadm/cli/notice.py", line 80, in confirm_prompt
    + "\n---\nSend now?"
TypeError: must be str, not NoneType
(synadm) ~/git/a-0_synadm (notice) $
(synadm) ~/git/a-0_synadm (notice) $
(synadm) ~/git/a-0_synadm (notice) $ synadm version
python_version: 3.6.9
server_version: 1.49.2

(synadm) ~/git/a-0_synadm (notice) $
```

synadm/cli/notice.py
```diff
@@ -0,0 +1,92 @@
...
+            for user in users:
+                if not re.search(to, user) is None:
+                    prompt = prompt + " - " + user + "\n"
+                    ctr = ctr + 1
+                    if ctr >= 10:
+                        prompt = prompt + " - ..."
+                        break
+            prompt = prompt + "\nUnformatted message:\n---\n" + plain_content\
+                + "\n---\nFormatted message:\n---\n" + formatted_content\
+                + "\n---\nSend now?"
```
JOJ0, line 80: A possible solution for above mentioned problem might be to, instead of concatenating strings with `+`, making use of f-strings (`f'Plaintext {variable}')` or a strings `format()` method. I think both would handle the type None by translating it it to the string "None", not sure if that would be the perfect solution here or the problem needs to be solved somewhere else.

JOJ0 (changes requested):

synadm/cli/notice.py
JOJ0: Update: notice API exists since Synapse 0.30, so that' not the issue ;-) I think it's the default value of the formatted_message that is None and thus, if not given, this concatenation error is thrown.

synadm/api.py
```diff
@@ -1139,3 +1139,44 @@ def user_shadow_ban(self, user_id, unban):
...
+            "content": {
+                "msgtype": "m.text",
+                "body": content_plain,
+                "format": "org.matrix.custom.html",
+                "formatted_body": content_html
+            }
+        }
+
+        # A regular expression was supplied to match receivers.
+        if receivers[:1] == '^':
```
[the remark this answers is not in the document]
Ascurius: In my experience, you use a `^` to indicate that you only want matches in a string that are found right at the beginning of that string, so you don't do any potentially incorrect multi-matching. 

But here the point is that, as you said, we can also have regex that doesn't use `^`. You are right that a distinction must be made here, but I find it not so easy to determine this on the basis of one or more characters, since these characters could potentially be part of a regex again. Spontaneously I would choose special characters, which are not allowed to be part of a MXID according to the matrix specification and at the same time can't be used for matching the allowed characters. However, I can't think of any character or character chain that would fulfill this condition.

Alternatively, a simple solution would be to add a flag for regex, with which the user can then explicitly indicate that the "search string" should be interpreted as regex. In fact I think this solution would be straight forward and one of the best. Let me know what you think about this proposal.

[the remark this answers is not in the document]
a-0-dev: @JOJ0 You're right, now I get what you mean.

@Ascurius I like your solution best, it's the most explicit and thus safe option to use imo. It makes the commands a little more verbose, but I think that's negligible wtih a short hand of the flag.

Maybe it would make sense to check for invalid characters in the `generate_mxid` function (in the else case where the mxid is constructed from a localpart)? So in case someone passes a regex without the flag, the function won't try to use the regex as a localpart... This could of course also be done in the CLI part, but I think hardening lib functions against arbitrary input is always a good idea. Let me know whether that makes sense :)

synadm/cli/notice.py
a-0-dev: I missed one case of input parameters: Do not read content from files, but no formatted body supplied via CLI. In that case formatted_content was set to None, which of course causes issues. Will be fixed in my next commit.

synadm/cli/notice.py
a-0-dev: Oh you found it yourself. Adding the extra edge case if-branch is necessary anyways because the formatted content variable must be set to send the message (this notice module will never send plain messages, if the user supplies only plain text it send a formatted message with plain content for both text variants). So that is the more robust solution I guess.

synadm/api.py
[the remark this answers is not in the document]
JOJ0: My approach/suggestion still would be (as discussed in #synadm:peek-a-boo.at) to immitate what Ansible dev's did: Use the tilde character (~) as a prefix for regex, since it doesnt have any special meaning in any regex dialect. https://docs.ansible.com/ansible/2.8/user_guide/intro_patterns.html

On the other hand I always welcome an explicit and straight forward approach like using an option. You decide @a-0-dev what you prefer. Both perfectly fine solutions IMO.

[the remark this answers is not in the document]
JOJ0: > Maybe it would make sense to check for invalid characters in the `generate_mxid` function (in the else case where the mxid is constructed from a localpart)? So in case someone passes a regex without the flag, the function won't try to use the regex as a localpart... This could of course also be done in the CLI part, but I think hardening lib functions against arbitrary input is always a good idea. Let me know whether that makes sense :)
> 

Yes very good idea, we should prevent that. I would like to have this change in a separate PR though. If you find the time that would be great, well, if not I'm being lazy and would also accept it within this PR ;-)😅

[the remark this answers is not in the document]
a-0-dev: I'll do this in a separate branch, but they should be accepted simultaneously because only having this PR merged without making `generate_mxid` more robust is prone to cause issues

JOJ0:
Thanks for those last changes, looking good and we are close to ready-to-merge. Some tiny formatting and help text notes though:

The META of the "Unformatted message" is PLAIN but the confirm_prompt tells "Unformatted message", this could be streamlined.

I suggest you remove the -a/--from-argument entirely, I think it's misleading what it does. I think your intention was that this boolean option has two states: Take from file or do the default: Take the cli args. I think we are better off without it. When --from-file is simply a flag, it is more clear what it's about.

Set show_defaults=True for the --paginate option.

Something is not working when paginate and regex is used together:
```
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send '.*user' "plain message for text users" '<strong>formatted msg</strong>' -r --paginate 3
Recipients (list may be incomplete):

Unformatted message:
---
plain message for text users
---
Formatted message:
---
<strong>formatted msg</strong>
---
Send now? [y/N]:   
```

It's fine when a matirx ID or just localpart is given and --to-regex is not present:
```
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send 'testuser1' "plain message for text users" '<strong>formatted msg</strong>' --paginate 1 
#Recipients (list may be incomplete):
 - @testuser1:peek-a-boo.at

Unformatted message:
---
plain message for text users
---
Formatted message:
---
<strong>formatted msg</strong>
---
Send now? [y/N]: 
```

JOJ0:
Found some formatting issues in notice.py. Trailing spaces and line continuation issues. You might want to run flake8 over this file as well., Thanks!

JOJ0:
Also I just realizse that we not display what the API returns. Please display it at the very end of the cli method using helper.output():

```
Once the notice has been sent, the API will return the following response:


{
    "event_id": "<event_id>"
}
```

a-0-dev:
> Thanks for those last changes, looking good and we are close to ready-to-merge. Some tiny formatting and help text notes though:
> 
> The META of the "Unformatted message" is PLAIN but the confirm_prompt tells "Unformatted message", this could be streamlined. I suggest using "Plaintext message" for the prompt (Or rather "Plain text message", not sure??)
> 
> Since FORMATTED only supports html, the brackets around HTML could be removed in the help text:
> 
> ```
> FORMATTED - HTML-formatted content of the notice. If not set, PLAIN will be used.
> ```

Done. Will be committed soon.

> I suggest you remove the -a/--from-argument entirely, I think it's misleading what it does. I think your intention was that this boolean option has two states: Take from file or do the default: Take the cli args. I think we are better off without it. When --from-file is simply a flag, it is more clear what it's about.

Done. Will be committed soon.

> Set show_defaults=True for the --paginate option. Also I wonder if streamlining with other synadm commands could be a good idea here. We have --limit in other places for pagination but I'm not entirely sure if it's exactly the same purpose as with this command. Have a look at `synadm room list --limit` for example to see what I'm talking about and help decide if it should be streamlined. Thanks!

show_defaults is done and will be committed soon. A limit is not pagination though, it only has similar effects at the beginning. Pagination does not limit anything, it just specifies how many users receive the notice at once, before synadm queries the next (`paginate` large) batch of users. Of course, in the first query it limits how many will be returned, but with pagination you also use the `next_token` (an offset so to say) to query the next consecutive batch.

So `limit` would be a false wording here, pagination is the correct term for this case. In room_list, `limit` is the correct wording.

> Something is not working when paginate and regex is used together:
> 
> ```
> (synadm) ~/git/a-0_synadm (notice) $ synadm notice send '.*user' "plain message for text users" '<strong>formatted msg</strong>' -r --paginate 1
> Recipients (list may be incomplete):
> 
> Unformatted message:
> ---
> plain message for text users
> ---
> Formatted message:
> ---
> <strong>formatted msg</strong>
> ---
> Send now? [y/N]:   
> ```
> 
> It's fine when a matirx ID or just localpart is given and --to-regex is not present:
> 
> ```
> (synadm) ~/git/a-0_synadm (notice) $ synadm notice send 'testuser1' "plain message for text users" '<strong>formatted msg</strong>' --paginate 1 
> #Recipients (list may be incomplete):
>  - @testuser1:peek-a-boo.at
> 
> Unformatted message:
> ---
> plain message for text users
> ---
> Formatted message:
> ---
> <strong>formatted msg</strong>
> ---
> Send now? [y/N]: 
> ```

The example receipients is just a "tiny bous feature" that is supposed to assist the sender, it only matches the users in the first batch and if the regex applies, shows them in a list. If no user matches the regex in the first batch (which is likely if you set `paginate` to 1), the list will be empty. I considered querying all users until the list has 10 entries or all users have been queried, but that would have inflated the CLI-side code so I stuck with the easier option.

To ease this issue, I could imagine not using `paginate` in this dry run, but a fixed constant like 100, to make it more likely that _some_ matrix id will be matched. `paginate` is a performance option anyways and should not accect the actual functionality of the command, so using a constant here sounds reasonable to me. What do you think?

JOJ0:
I dont understand. In my first example exactly one user named testuser1 should be shown with paginate set to 1. That not what's expected?

JOJ0:
Give it a try with the constant and I'll see how it feels then from a user perspective.

a-0-dev:
I set it to 100 now, you are probably not going to feel anything now, it will feel alright.

The point where this constant 100 feels awkward is when your server has a larger (e.g. >200) number of users and you apply a fairly restrictive regex. That way there may not be any match in the first 100, even though there are matches in later pages of users. 

From a user perspective it would definitely make the most sense to search for regex matches until you have 10 matches or you searched all users. We would need to accept a larger `notice_send_cmd` then though, since this means quite some more logic. I currently have some time on my hands, so I'll do a demo-implementation and see how bad it actually would be.

a-0-dev:
As promised, here it is, and it has become a bigger function. However, I must admit I like this version a lot more as it feels more solid.

The user now has the guarantee that only the users in the "dry run list" will receive the notice - and the user can specify the upper bound of that list via CLI option. So if you really want to, you can check all 200 or 300 mxids yourself before sending. 

The list ends with ` - ...` if there is at least one mxid not mentioned in the list (limit too small) and if this last line is missing you _know_ these are the only mxids.

So I'd like to keep this.

JOJ0:
Yes this feels right now! Very nice! ... is clearly showing that the list is incomplete! Great!

 --match-list-length is a rather longish name but at least it's descriptive. I don't have a better idea. I'll think about it.
Maybe you wanna put short forms for this option as well as for --paginate? -l and -p for example?

To make flake8 happy with the @options definitions you would want to use this line-continuation syntax (line-break must be directly after the opening bracket!):

```
click.option(
    "--from-file", "-f", default=False, show_default=True,
    is_flag=True, help="""Interpret arguments as file paths instead of notice
    content and read content from those files""")
```
HTH

a-0-dev:
I first didn't like the idea of using `-l` because it very much sounded like `--limit` to me, but to be honest it's still a nice shorthand for a *L*ist *L*ength, and "limit" is also half-true'ish. And you shouldn't use unknown options without looking at the help page anyways ;) So `-l` is the way to go I guess.

JOJ0:
My typing errors around awkward .... syntax are actually very spot on the topic! 🤣

JOJ0 (approved):

JOJ0:
I approved the last changes now but might start a new review and change request.

synadm/cli/notice.py
```diff
@@ -0,0 +1,124 @@
...
+@click.option(
+    "--paginate", type=int, default=100, show_default=True, help="""Sets how
+    many users will be retrieved from the server at once""")
+@click.option(
+    "--to-regex", "-r", default=False, show_default=True, is_flag=True,
+    help="Interpret TO as regular expression")
+@click.option(
+    "--match-list-length", "-l", type=int, default=10, show_default=True,
+    help="""Length of the displayed list of matched receivers. Does not impact
+    sending behavior""")
```
JOJ0, line 47: Also here I think there is just a little room for improvement. It should be made clear that this option is about the recipient list that is only shown in the dry run. It kind of tells that but it could be made even better. Maybe just adding some to the first sentence would do the trick:

```
Length of the displayed list of potential recipients shown in the confirmation prompt.
```

```
Length of the displayed list of matched recipients shown in the confirmation prompt.
```

```
Length of the displayed list of recipients shown in the confirmation prompt.
```

```
How many recipients should be shown in the confirmation prompt?
```

```
How many recipients should be shown in the message preview?
```

Just brainstorming here, might not all be perfect...

JOJ0: Also something like this could be added for exactly documenting what this option is about and what it is not:

```
This option is ignored when the global `--batch` flag is set, since there won't be a confirmation prompt. 
```

JOJ0 (changes requested):
Especially I'm picky with the help texts now. This is the final polishing for a well-rounded command IMO. Document verbosely what the command does and there won't be surprises to the user (or at least no excuses ;-)) Thanks for your patience! Appreciated!

synadm/cli/notice.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@notice.command(name="send")
+@click.option(
+    "--from-file", "-f", default=False, show_default=True, is_flag=True,
+    help="""Interpret arguments as file paths instead of notice content and
+    read content from those files""")
+@click.option(
+    "--paginate", type=int, default=100, show_default=True, help="""Sets how
+    many users will be retrieved from the server at once""")
```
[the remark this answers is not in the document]
JOJ0: Also a shorthand option would be nice here. Why not just `-p`?

synadm/cli/notice.py
a-0-dev: I included the second suggestion you made, it's a good improvement. However, I think adding this (explicitly stating it is about the confirmation prompt) makes it clear enough that it won't have any effect if you switch off the confirmation prompt. I would argue in favor of brievety here.

a-0-dev:
Since the regex part became quite vital when testing the send command, I decided to include the few changes of the `generate_mxid` regex here instead of a separate pull request. (commit [a700ec4](https://github.com/JOJ0/synadm/pull/74/commits/a700ec4ba65bec434387466a3d73cff0fa29c12f) )

A noticable difference however is that the "replace @ and :" part is gone now (was your goal to convert `@user:` to `user` to `@user:retrieved.hs.tld`?). We can change that part back if you wish, to me it made most sense to match on None, then full valid mxid, then valid localpart, then return None since nothing is matched.

Imo (but this is very opinionated) you should not accept "too wrong inputs" as it becomes unclear to the user what actually happens. And I don't feel like the ability to pass @username instead of username is an actual improvement. Either mxid, or localpart, or regex. But no character manipulation to "make it fit". Just my opinion though, maybe you have different thoughts?

JOJ0 (approved):

JOJ0:
> Since the regex part became quite vital when testing the send command, I decided to include the few changes of the `generate_mxid` regex here instead of a separate pull request. (commit [a700ec4](https://github.com/JOJ0/synadm/pull/74/commits/a700ec4ba65bec434387466a3d73cff0fa29c12f) )

That's ok and appreciated!

> A noticable difference however is that the "replace @ and :" part is gone now (was your goal to convert `@user:` to `user` to `@user:retrieved.hs.tld`?). We can change that part back if you wish, to me it made most sense to match on None, then full valid mxid, then valid localpart, then return None since nothing is matched.
> 
> Imo (but this is very opinionated) you should not accept "too wrong inputs" as it becomes unclear to the user what actually happens. And I don't feel like the ability to pass @username instead of username is an actual improvement. Either mxid, or localpart, or regex. But no character manipulation to "make it fit". Just my opinion though, maybe you have different thoughts?

Actually this behaviour was intentionally. @user or user: was just accepted because it is clearly not an mixd.

I am fine with your decision but other problems pop up when I try it out now. It seems it is sent to Synapse now if I put @user or @user: 

```
(synadm) ~/git/a-0_synadm (notice) $ synadm user details @testuser1
errcode: M_INVALID_PARAM
error: Expected UserID string to start with '@'


(synadm) ~/git/a-0_synadm (notice) $ synadm user details @testuser1:
errcode: M_INVALID_PARAM
error: Expected UserID string to start with '@'


(synadm) ~/git/a-0_synadm (notice) $ synadm user details testuser1:
errcode: M_INVALID_PARAM
error: Expected UserID string to start with '@'
````

Shouldn't something like this catched before and not even sent to Synapse? It's misleading: It tells  the user he must put in a full mxid, which is not true!

**Comment by JOJ0** [comment]
Either you just keep the basic logic of that condition block  or you now test through all the other commands that rely on that regex function's behaviour. Each and everything that's handling something with usernames is relying on that.

I suggest you remove the second elif branch, so we keep close to the original behaviour.

What we could do that can't hurt: In the upgraded regex we now have, please put a ^ sign at the very beginning: `^@[....`

**Comment by JOJ0** [comment]
I think this works well:

```
        if user_id is None:
            return None
        elif re.match(r"^@[-./=\w]+:[-\[\].:\w]+$", user_id):
            return user_id
        else:
            localpart = re.sub("[@:]", "", user_id)
            mxid = "@{}:{}".format(localpart, self.retrieve_homeserver_name())
            return mxid

```

I just tested with some `synadm user ...` commands

**Comment by a-0-dev** [comment]
> I am fine with your decision but other problems pop up when I try it out now. It seems it is sent to Synapse now if I put @user or @user:

> Shouldn't something like this be catched before and not even sent to Synapse? It's misleading: It tells the user he must put in a full mxid, which is not true!

Well, the issue is that `generate_mxid` before and now _could_ return `None`, but that case is generally not handled in synadm (except for my code :P), with my changes it would return `None` for something like @user , which is where the problems start.

> I think this works well:
```
        if user_id is None:
            return None
        elif re.match(r"^@[-./=\w]+:[-\[\].:\w]+$", user_id):
            return user_id
        else:
            localpart = re.sub("[@:]", "", user_id)
            mxid = "@{}:{}".format(localpart, self.retrieve_homeserver_name())
            return mxid
```
This is close to what we had before, with the major problem why we started discussing changes: If someone passes ".*" as TO (in `notice send`), but forgot to set the regex flag, it is neither `None`, nor a valid mxid, so it lands in your `else` case and is force-built into some mxid which doesn't make any sense - and subsequently processed by synapse because `generate_mxid` didn't return None. That is what I wanted to avoid...

> What we could do that can't hurt: In the upgraded regex, please put a ^ sign at the very beginning: `^@[....`

This is not necessary when using `re.match`, we could still do it to improve code readability. See https://docs.python.org/3/library/re.html#re.match

**Comment by JOJ0** [comment]
Ok thanks for the reminder where we are coming from. For the user commands, this is how it reacts:
```
~/git/a-0_synadm (notice) $ synadm user details '.*'
WARNING Synapse returned status code 404
errcode: M_NOT_FOUND
error: User not found

~/git/a-0_synadm (notice) $ 
~/git/a-0_synadm (notice) $ 
~/git/a-0_synadm (notice) $ 
~/git/a-0_synadm (notice) $ synadm user details '....'
WARNING Synapse returned status code 404
errcode: M_NOT_FOUND
error: User not found

~/git/a-0_synadm (notice) $ synadm user details '@.*'
WARNING Synapse returned status code 404
errcode: M_NOT_FOUND
error: User not found

~/git/a-0_synadm (notice) $ synadm user details '@.*:bla.de'
WARNING Synapse returned status code 404
errcode: M_NOT_FOUND
error: User not found

~/git/a-0_synadm (notice) $ 
```

Not too bad and IMHO absolutely ok like that!

Let me see what would happen with the notice send command.....

**Comment by JOJ0** [comment]
Well, not too bad either:
```
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send '.*' "plain message for text users" '<strong>formatted msg</strong>'
Recipients:
 - @.*:peek-a-boo.at

Plaintext message:
---
plain message for text users
---
Formatted message:
---
<strong>formatted msg</strong>
---
Send now? [y/N]: y
WARNING Synapse returned status code 404
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send '.....' "plain message for text users" '<strong>formatted msg</strong>'
Recipients:
 - @.....:peek-a-boo.at

Plaintext message:
---
plain message for text users
---
Formatted message:
---
<strong>formatted msg</strong>
---
Send now? [y/N]: y
WARNING Synapse returned status code 404
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send '@.*' "plain message for text users" '<strong>formatted msg</strong>'
Recipients:
 - @.*:peek-a-boo.at

Plaintext message:
---
plain message for text users
---
Formatted message:
---
<strong>formatted msg</strong>
---
Send now? [y/N]: y
WARNING Synapse returned status code 404
(synadm) ~/git/a-0_synadm (notice) $ 
```

ok it's a 404, not tooo nice of a response. Can you please test how it reacts with your proposal?

Too me it looks like silently ignoring stuff. This is what I get with your extra `elif` branch:


```
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send '@.*' "plain message for text users" '<strong>formatted msg</strong>'
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send '.*' "plain message for text users" '<strong>formatted msg</strong>'
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send '' "plain message for text users" '<strong>formatted msg</strong>'
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send blabla "plain message for text users" '<strong>formatted msg</strong>'
Recipients:
 - @blabla:peek-a-boo.at

Plaintext message:
---
plain message for text users
---
Formatted message:
---
<strong>formatted msg</strong>
---
Send now? [y/N]: y
WARNING Synapse returned status code 404
(synadm) ~/git/a-0_synadm (notice) $ 

```

It is not perfect either I think.....could be better....any error in this case is better than silently ignoring (from user perspective)

**Comment by a-0-dev** [comment]
I agree silently ignoring is not too great, I could add a custom error output in case `generate_mxid` returns `None` (that check is done already, it's just adding one `print(...)` statement).

However I would generally argue we should not send mxids to synapse of which we could already know they don't exist (or contain arbitrary characters like `[` or `*`), since we can control how our program behaves but passing such stuff to synapse may cause unwanted side effects

**Comment by JOJ0** [comment]
> > What we could do that can't hurt: In the upgraded regex, please put a ^ sign at the very beginning: ^@[....
> 
> This is not necessary when using re.match, we could still do it to improve code readability. See https://docs.python.org/3/library/re.html#re.match

Yes definitely! Just for readability. You mentioned the improvement of using re.match over re.search. :+1:

**Comment by JOJ0** [comment]
> I agree silently ignoring is not too great, I could add a custom error output in case `generate_mxid` returns `None` (that check is done already, it's just adding one `print(...)` statement).

That sounds right!

> However I would generally argue we should not send mxids to synapse of which we could already know they don't exist (or contain arbitrary characters like `[` or `*`), since we can control how our program behaves but passing such stuff to synapse may cause unwanted side effects

That's certainly correct and would be the best of situations, but still Synapse is aware of handling errors itself and giving useful error responses. Not always, but often.

But yes I totaly agree that reacting on gereate_mxid returning None is worth a try! Go for it and then let's test again!

**Comment by JOJ0** [comment]
I was testing your last change yesterday and it still didnt feel well rounded. Sometimes we are getting an unknown recepient error and sometmes not. It feels random, thus the following idea coming up in a band rehearsal break ;-)

Try removing all the magic: No substitution of @ and :, as you already suggested previously, remove check for forgotten --regex.

Just an idea and didnt think it through thoroughly!!! Test how it behaves with user details command as well! Post some testing examples.

HTH

File: synadm/api.py [hunk 15]
```diff
@@ -1139,3 +1140,47 @@ def user_shadow_ban(self, user_id, unban):
...
+                "body": content_plain,
+                "format": "org.matrix.custom.html",
+                "formatted_body": content_html
+            }
+        }
+
+        # A regular expression was supplied to match receivers.
+        if to_regex:
+            outputs = []
+            response = self.user_list(0, paginate, True, False, "", "")
```
JOJ0, line 1166: Also I was thinking: We are doing this only when regex option given, why not always check if the user is existing before showing the confirmation prompt? Performance reasons?

JOJ0:
I put some thought and testing into this and came up with the following:

```
    def generate_mxid(self, user_id):
        """ Checks whether the given user ID is an MXID already or else
        generates it from the passed string and the homeserver name fetched
        via the retrieve_homeserver_name method.

        Args:
            user_id (string): User ID given by user as command argument.

        Returns:
            string: the fully qualified Matrix User ID (MXID) or None if the
                user_id parameter is None or no regex matched.
        """
        if user_id is None:
            self.log.debug("Missing input in generate_mxid().")
            return None
        elif re.match(r"^@[-./=\w]+:[-\[\].:\w]+$", user_id):
            self.log.debug("A proper MXID was passed.")
            return user_id
        elif re.match(r"^@?[-./=\w]+:?$", user_id):
            self.log.debug("A proper localpart was passed, generating MXID "
                           "for local homeserver.")
            localpart = re.sub("[@:]", "", user_id)
            mxid = "@{}:{}".format(localpart, self.retrieve_homeserver_name())
            return mxid
        else:
            self.log.error("Neither an MXID nor a proper localpart was "
                           "passed.")
            return None
```

Also I realized that the only way to not screw things up with core changes like this is to be very explicit, log properly what's happening and most importantly throw an error when things definitely won't work out for the caller!

Hope that helps! Please do a lot of testing and post odd things!

One thing that is odd but now I think I can live with since the first error message gives us a clue where the problem is:
```
(synadm) ~/git/a-0_synadm (notice) $ synadm user details 'testuser::'
ERROR Neither an MXID nor a proper localpart was passed.
errcode: M_INVALID_PARAM
error: Expected UserID string to start with '@'

(synadm) ~/git/a-0_synadm (notice) $
```

In the case of the notice send command this is behaving very well since we're reacting on the None returned already:
```
(synadm) ~/git/a-0_synadm (notice) $ synadm -vv notice send 'test::' "plain msg" 
DEBUG Config entry read. user: admin
DEBUG Config entry read. token: REDACTED
DEBUG Config entry read. base_url: https://peek-a-boo.at
DEBUG Config entry read. admin_path: /_synapse/admin
DEBUG Config entry read. matrix_path: /_matrix
DEBUG Config entry read. timeout: 3600
DEBUG Config entry read. server_discovery: well-known
DEBUG Config entry read. homeserver: peek-a-boo.at
DEBUG Config entry read. format: yaml
DEBUG Formatter in use: yaml - <function dump at 0x7f613d6e9510>
ERROR Neither an MXID nor a proper localpart was passed.
The recipient you specified is invalid.
(synadm) ~/git/a-0_synadm (notice) $ 
```

synadm/api.py
a-0-dev: I don't think there was a specific reason for me to not do it for single recipients. I would suggest to change the `else` branch to the following:
```
else:
            data["user_id"] = generate_mxid(receivers)
            return [self.query("post", f"v1/send_server_notice", data=data)]
```
In my opinion checking whether the user exists is not necessary in our code as that is really something synapse can do best and gives good error messages for. But the "input checking" part (illegal characters, generating mxid from localpart) can and should be done by synadm - and it's achieved by this minor change.

a-0-dev:
> I put some thought and testing into this and came up with the following:
> 
> ```
>     def generate_mxid(self, user_id):
>         """ Checks whether the given user ID is an MXID already or else
>         generates it from the passed string and the homeserver name fetched
>         via the retrieve_homeserver_name method.
> 
>         Args:
>             user_id (string): User ID given by user as command argument.
> 
>         Returns:
>             string: the fully qualified Matrix User ID (MXID) or None if the
>                 user_id parameter is None or no regex matched.
>         """
>         if user_id is None:
>             self.log.debug("Missing input in generate_mxid().")
>             return None
>         elif re.match(r"^@[-./=\w]+:[-\[\].:\w]+$", user_id):
>             self.log.debug("A proper MXID was passed.")
>             return user_id
>         elif re.match(r"^@?[-./=\w]+:?$", user_id):
>             self.log.debug("A proper localpart was passed, generating MXID "
>                            "for local homeserver.")
>             localpart = re.sub("[@:]", "", user_id)
>             mxid = "@{}:{}".format(localpart, self.retrieve_homeserver_name())
>             return mxid
>         else:
>             self.log.error("Neither an MXID nor a proper localpart was "
>                            "passed.")
>             return None
> ```
> 
> Also I realized that the only way to not screw things up with core changes like this is to be very explicit, log properly what's happening and most importantly throw an error when things definitely won't work out for the caller!
> 
> Hope that helps! Please do a lot of testing and post odd things!
> 
> One thing that is odd but now I think I can live with since the first error message gives us a clue where the problem is:
> 
> ```
> (synadm) ~/git/a-0_synadm (notice) $ synadm user details 'testuser::'
> ERROR Neither an MXID nor a proper localpart was passed.
> WARNING Synapse returned status code 400
> errcode: M_INVALID_PARAM
> error: Expected UserID string to start with '@'
> 
> (synadm) ~/git/a-0_synadm (notice) $
> ```
> 
> In the case of the notice send command this is behaving very well since we're reacting on the None returned already:
> 
> ```
> (synadm) ~/git/a-0_synadm (notice) $ synadm notice send 'test::' "plain msg" 
> ERROR Neither an MXID nor a proper localpart was passed.
> The recipient you specified is invalid.
> (synadm) ~/git/a-0_synadm (notice) $ 
> ```
> 
> So actually this should be easy to fix within all the commands that handle users! I'll take of that myself later. Don't worry.
> 
> FYI: To see the log.debug messages while testing you need to pass `-vv` to the main `synadm` command.
> 
> Also note that I adapted the Returns: docstring to fit the new behaviour.

Seems to work great with the inputs I tested, I'll commit this version of `generate_mxid`. I also think handling the `None` output of this function is very necessary anyways, so it's good that we stumbled across it now

synadm/api.py
JOJ0: Can't work like that. generate_mxid() is not available in SynapseAdmin object. You are alreading passing through it in notice.py. Need to revert this commit: https://github.com/JOJ0/synadm/pull/74/commits/3f8b78e5cbd40213d3f60c0a7d64032e389276ae

JOJ0:
Still something in your cojnfirm_prompt functrion needs to be improved. It just doesn't feel correct with some crap inputs that don't match anything. Something like: No users matching the regex pattern.

```
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send '@testuser.*' "plain msg" -r
Recipients:
 - @testuser1:peek-a-boo.at
 - @testuser2:peek-a-boo.at

Plaintext message:
---
plain msg
---
Formatted message:
---
plain msg
---
Send now? [y/N]: n
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send 'testuser.*' "plain msg" -r
Recipients:

Plaintext message:
---
plain msg
---
Formatted message:
---
plain msg
---
Send now? [y/N]: 
(synadm) ~/git/a-0_synadm (notice) $ 
(synadm) ~/git/a-0_synadm (notice) $ synadm notice send 'testuser.*' "plain msg" -r
Recipients:

Plaintext message:
---
plain msg
---
Formatted message:
---
plain msg
---
Send now? [y/N]: y
(synadm) ~/git/a-0_synadm (notice) $ 
```

a-0-dev:
> Still something in your cojnfirm_prompt functrion needs to be improved. It just doesn't feel correct with some crap inputs that don't match anything. Something like: No users matching the regex pattern.
> 
> ```
> (synadm) ~/git/a-0_synadm (notice) $ synadm notice send '@testuser.*' "plain msg" -r
> Recipients:
>  - @testuser1:peek-a-boo.at
>  - @testuser2:peek-a-boo.at
> 
> Plaintext message:
> ---
> plain msg
> ---
> Formatted message:
> ---
> plain msg
> ---
> Send now? [y/N]: n
> (synadm) ~/git/a-0_synadm (notice) $ 
> (synadm) ~/git/a-0_synadm (notice) $ 
> (synadm) ~/git/a-0_synadm (notice) $ 
> (synadm) ~/git/a-0_synadm (notice) $ synadm notice send 'testuser.*' "plain msg" -r
> Recipients:
> 
> Plaintext message:
> ---
> plain msg
> ---
> Formatted message:
> ---
> plain msg
> ---
> Send now? [y/N]: 
> (synadm) ~/git/a-0_synadm (notice) $ 
> ```

Hmmm that is true, but what would you expect? Output "No user matched the regular expression" and quit?

> Also when trying to send to nothing it just doesn't error out and just does nothing.
> 
> ```
> (synadm) ~/git/a-0_synadm (notice) $ synadm notice send 'testuser.*' "plain msg" -r
> Recipients:
> 
> Plaintext message:
> ---
> plain msg
> ---
> Formatted message:
> ---
> plain msg
> ---
> Send now? [y/N]: y
> (synadm) ~/git/a-0_synadm (notice) $ 
> ```
> 
> It's not clear to the user what's happening....

I mean for the --batch version, that is absolutely what's supposed to happen. Send to all matching users and quit afterwards. As for the UI, I agree the empty receipients list doesn't look like a list in the console, so it may not be clear that it actually says "no recipient matched"...

Maybe I should just add "no recipient matched" below `Recipients:` if no mxid was printed during iterating? That would make the output a lot more clear without major changes in how the code works and without inflating the code (as would be necessary if it was to be checked whether any recipient matched *before* starting to build the output string)

a-0-dev:
Seems to work just as I would expect, thanks! I'm sorry for bringing up two more topics though :)
1. I'm not happy with my own wording of the regex option any more. "--to-regex" _very much_ sounds like synadm would convert something to a regex, which is very misleading (at least it doesn't make any sense, so the user would probably think twice and come to the right conclusion). Maybe rename it to "--regex-to" before the release so it's not changed later on? It sounds a little off/wrong, but the only way to understand it is that TO is a regex - so imho a little more clear and thus better. What do you think?
2. I built this in a very scalable way for a reason, it's supposed to be able to send notices to thousands / 10.000s of users without any issue. I identified a potential bottleneck when I first saw the console output of event IDs now (which I like a lot!). It showed me that the `outputs` list (api.py L1177) will have as many entries as there are recipients - AND all of it will be printed to the console. Sure, not every server is matrix.org, but I think this may cause issues... What's your opinion? Not collecting the outputs at all? Only collecting some? Another option?

It would be cool to figure out point 1 before the release, point 2 isn't nearly as critical imo. Apart from that, I'd be happy to see the next version of synadm now :)

JOJ0:
1) I would radically shorten it to just --regex. We shouldn't forget that a user when looking at --help would instantly see the description next do it.

Also in the html rendered docs it would be like that. 

What I would also like about that,is in a commandline where the full option is in use, it's just quick to type.

JOJ0:
2 )

They will be printed all at once right? Yeah that could ptoblematic. Printing after each send would be nice actually but that could be difficult to realise.

Maybe introduce a --silent option?

a-0-dev:
> 1 ) I would radically shorten it to just --regex. We shouldn't forget that a user when looking at --help would instantly see the description next do it.
> 
> Also in the html rendered docs it would be like that.
> 
> What I would also like about that,is in a commandline where the full option is in use, it's just quick to type (I personally often use long options when using commands I'm not very familiar with, It helps reading shell history as well).

alright, good suggestion. Will commit that in a second.

JOJ0:
3 )
Don't be sorry! I'd like to thank you again for your extreme patience with getting this feature production-ready and even for still bringing up ideas and improvement suggestions. It was not an easy one but it was all worth the hassle in the end. I'm glad we made it this far :-)

a-0-dev:
> 2 )
> 
> They will be printed all at once right? Yeah that could ptoblematic. Printing after each send would be nice actually but that could be difficult to realise.
> 
> Maybe introduce a --silent option?

Not sure... maybe rather opt-in than opt-out? As an admin you probably don't know about this issue, and if you send a message and then realize you have this issue, it's already too late... And the use of the event IDs is rather limited imo (?)

Or include a warning in the confirmation prompt if `--silent` is not specified? That way we could keep the default-output (which may be important for consistency between commands) and the admins would know they may need to adjust their command before proceeding

JOJ0:
> alright, good suggestion. Will commit that in a second.

pr is merged already, I wont see it. But dont worry I'll do it soonish.

JOJ0:
> > 2 )
> > They will be printed all at once right? Yeah that could ptoblematic. Printing after each send would be nice actually but that could be difficult to realise.
> > Maybe introduce a --silent option?
> 
> Not sure... maybe rather opt-in than opt-out? As an admin you probably don't know about this issue, and if you send a message and then realize you have this issue, it's already too late... And the use of the event IDs is rather limited imo (?)
> 
> Or include a warning in the confirmation prompt if `--silent` is not specified? That way we could keep the default-output (which may be important for consistency between commands) and the admins would know they may need to adjust their command before proceeding

Hi, sorry for the delay. I realised there is no error handling for the reading from file features. I added it. 

Also I added a --silent option and a warning/note. At first I thought about adding it only if there is a rather large amount of recipients, let's say if its more than a hundred. I was wondering if there is any place in the code where this is actually known. I might have missed this place. In case you know one, please help me out, if there is none, no worries, let's just release it like it is. It's fine :-)

Have a look at my latest commits @a-0-dev. Thanks for any thoughts in advance!

a-0-dev:
> > > 2 )
> > > They will be printed all at once right? Yeah that could ptoblematic. Printing after each send would be nice actually but that could be difficult to realise.
> > > Maybe introduce a --silent option?
> > 
> > 
> > Not sure... maybe rather opt-in than opt-out? As an admin you probably don't know about this issue, and if you send a message and then realize you have this issue, it's already too late... And the use of the event IDs is rather limited imo (?)
> > Or include a warning in the confirmation prompt if `--silent` is not specified? That way we could keep the default-output (which may be important for consistency between commands) and the admins would know they may need to adjust their command before proceeding
> 
> Hi, sorry for the delay. I realised there is no error handling for the reading from file features. I added it.
> 
> Also I added a --silent option and a warning/note. At first I thought about printinhg it only if there is a rather large amount of recipients detected, let's say if its more than a hundred. I was wondering if there is any place in the code where this is actually known. I might have missed this place. In case you know one, please help me out, if there is none, no worries, let's just release it like it is. It's fine :-)
> 
> Have a look at my latest commits @a-0-dev. Thanks for any thoughts in advance!

I thought about this too (only printing the warning if there are many recipients) - but it is indeed unknown how many recipients there are in total (after all, this number is only known once synadm queried every mxid and matched the regex). Somehow depending on the pagination limit, which a good admin will probably have set appropriately, would probably be a fix, but a rather dirty one.

So I would suggest to leave it as-is, the warning is fine and the `--silent` option does its job. Thanks!

PR #77, opened 2022-09-24
Add aliases to --batch global flag and

JOJ0:
describe more precicesly what it does.


JOJ0:
@a-0-dev, since we were talking about this option, your review kindly requested! Thanks :-)

JOJ0:
I do agree that adding --yes to the list is fine and as you put it, nice because other tools use it, it's quite common!

But if we are being picky we could argue that switching off translation of unix timestamps to human readable is not what people would expect from a --yes option...

JOJ0:
Anyway, I merge this now since it's documented clearly that --yes does a little more than saying "yes" ;-) It's documented, it's fine :-)

PR #78, opened 2022-09-29
Linting fixes with the help of flake8

JOJ0:
Working towards a clean code base to be ready for #76

PR #80, opened 2022-11-21
Add flake8 gh-actions workflow

JOJ0:
Added a flake8 linting workflow using https://github.com/marketplace/actions/python-flake8-lint

Exclusions to what is being linted I previously added inline with the code in PR #78, except the whole file setup.py I excluded in lint.yaml.


Ascurius (approved):
Looks great!

PR #81, opened 2022-11-25
add command 3pid for finding user by their 3PID

ashfame:
This PR adds support for 3PID endpoint for finding users by their 3PID value which could be an `email` or `msidn` (specified as medium).

Please note this requires Synapse v1.72.0 (I currently haven't been able to test this as our EMS server is still v1.71.0 and would get upgraded hopefully next week but this is relatively straight forward).

Unlike other endpoints but similar to auth-provider endpoint, it requires 2 values:

- medium (`email` or `msidn`)
- address (actual email address or phone number as per medium specified)

> [Documentation for this admin api endpoint](https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#find-a-user-based-on-their-third-party-id-threepid-or-3pid)

#### Correct usage
```
$ synadm user 3pid -m email universe@github.com
{"user_id": "@ashfame:synapse.dev"}
```

Looking forward to your feedback! :)

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -670,6 +670,12 @@ class SynapseAdmin(ApiRequest):
         return self.query("get",
                           f"v1/auth_providers/{provider}/users/{external_id}")
 
+    def user_3pid_search(self, medium, address):
+        """ Finds a user based on their Third Party ID by specifying what kind
+        of 3PID it is as medium.
+        """
+        return self.query("get", f"v1/threepid/{medium}/users/{address}")
+
     def room_join(self, room_id_or_alias, user_id):
         """ Allow an administrator to join an user account with a given user_id
         to a room with a given room_id_or_alias
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -623,3 +623,20 @@ def auth_provider_search(helper, provider, external_user_id):
     helper.output(
         helper.api.user_auth_provider_search(provider, external_user_id)
     )
+
+
+@user.command(name="3pid")
+@click.argument("address", type=str)
+@click.option(
+    "-m", "--medium", is_flag=False, required=True, default="email",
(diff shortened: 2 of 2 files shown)


JOJ0:
I can't test it either since my Synapse is too old as well but as you say it's pretty straightforward. I'll merge it now  but will wait for your tests before I'll approach a next synadm release. Also if you feel like it, let's discuss in #synadm if it makes sense to include convenience commands into `synadm user search` with those last 2 PR's features.

Also, one thing I was wondering is if it would make sense to build a Synapse version check into some of these "newer" synadm commands. The  version apit method could be used for that: https://github.com/JOJ0/synadm/blob/master/synadm/api.py#L923-L926, but that shouldn't bother this PR and could be introduced for several commands in a sperate submission.

ashfame:
> I can't test it either since my Synapse is too old as well but as you say it's pretty straightforward. I'll merge it now but will wait for your tests before I'll approach a next synadm release.

Sure thing! I will ping you here once I am able to test it and confirm.

> Also if you feel like it, let's discuss in #synadm if it makes sense to include convenience commands into synadm user search with those last 2 PR's features.

Might be better done in a separate Github issue so that notifications and follow ups are a little easier? I don't have any urgency for it though. Also, haven't evaluated how it would make sense under the search command as of now.

> A further thing I was wondering about is if it would make sense to build a Synapse version check into some of these "newer" synadm commands. The version api method could be used for that: https://github.com/JOJ0/synadm/blob/master/synadm/api.py#L923-L926, but that shouldn't bother this PR and could be introduced for several commands in a sperate submission.

Sounds like that would be a nice enhancement. My thoughts: The way I think it would make sense to implement is, such commands could defined upwards what version of Synapse is supported and then whenever an API hits a 404, we check whether the current version is less than specified Synapse version for that command and display a message to the user. That way, its not slowing down the command since it requires checking version every time its invoked.

PR #82, opened 2022-11-28
Correction: use the right function

ashfame:
PR fixes the correct function name to be invoked in 3pid command added in last PR (#81).

Not sure what happened but feels like deja vu. I made this change earlier as well 🤷 

Sorry about this. I have now tested this with Synapse v1.72.0 and it works fine.


ashfame:
Also, if you can push a new release out, that would be great. This will enable me to use the 3PID command internally :)

PR #88, opened 2023-02-02
Add ssl_verify to manage selfsigned certificate

heurtematte:
Allow to param ssl_verify in configuration, allow to request https localhost homeserver with self signed certificate.

`~/.config/synadm.yaml`

```
ssl_verify: False
```

Signed-off-by: sebastien.heurtematte <sebastien.heurtematte@eclipse-foundation.org>

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -57,7 +57,7 @@ class ApiRequest:
             timeout (int): requests module timeout used in query method
             debug (bool): enable/disable debugging in requests module
             verify(bool): SSL verification is turned on by default
-                and can be turned off using this method.
+                and can be turned off using this argument.
         """
         self.log = log
         self.user = user
@@ -389,7 +389,7 @@ class SynapseAdmin(ApiRequest):
                 method
             debug (bool): enable/disable debugging in requests module
             verify(bool): SSL verification is turned on by default
-                and can be turned off using this method.
+                and can be turned off using this argument.
         """
         super().__init__(
             log, user, token,
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -41,7 +41,8 @@ class ApiRequest:
 
     This is subclassed by SynapseAdmin and Matrix
     """
-    def __init__(self, log, user, token, base_url, path, timeout, debug):
+    def __init__(self, log, user, token, base_url, path, timeout, debug,
+                 verify=None):
         """Initialize an APIRequest object
 
         Args:
@@ -55,6 +56,8 @@ class ApiRequest:
                 base_url to form the basis for all API endpoint paths
             timeout (int): requests module timeout used in query method
             debug (bool): enable/disable debugging in requests module
+            verify(bool): SSL verification is turned on by default
+                and can be turned off using this method.
         """
         self.log = log
         self.user = user
@@ -68,9 +71,10 @@ class ApiRequest:
         self.timeout = timeout
         if debug:
             HTTPConnection.debuglevel = 1
+        self.verify = verify
 
     def query(self, method, urlpart, params=None, data=None, token=None,
-              base_url_override=None, verify=True):
+              base_url_override=None, verify=None):
         """Generic wrapper around requests methods.
 
         Handles requests methods, logging and exceptions.
@@ -108,10 +112,14 @@ class ApiRequest:
             self.log.debug("Token override! Adjusting headers.")
             self.headers["Authorization"] = "Bearer " + token
 
+        override_verify = self.verify
+        if verify is not None:
+            override_verify = verify
+
         try:
             resp = getattr(requests, method)(
                 url, headers=self.headers, timeout=self.timeout,
-                params=params, json=data, verify=verify
+                params=params, json=data, verify=override_verify
             )
             if not resp.ok:
                 self.log.warning(f"{host_descr} returned status code "
@@ -196,7 +204,7 @@ class MiscRequest(ApiRequest):
         ApiRequest (object): parent class containing general properties and
             methods for requesting REST API's
     """
-    def __init__(self, log, timeout, debug):
+    def __init__(self, log, timeout, debug, verify=None):
         """Initialize the MiscRequest object
 
         Args:
@@ -204,11 +212,13 @@ class MiscRequest(ApiRequest):
             timeout (int): requests module timeout used in ApiRequest.query
                 method
             debug (bool): enable/disable debugging in requests module
+            verify(bool): SSL verification is turned on by default
+                and can be turned off using this method.
         """
         super().__init__(
             log, "", "",  # Set user and token to empty string
             "", "",  # Set base_url and path to empty string
-            timeout, debug
+            timeout, debug, verify
         )
 
     def federation_uri_well_known(self, base_url):
@@ -243,7 +253,7 @@ class Matrix(ApiRequest):
             methods for requesting REST API's
     """
     def __init__(self, log, user, token, base_url, matrix_path,
-                 timeout, debug):
+                 timeout, debug, verify):
         """Initialize the Matrix API object
 
         Args:
@@ -258,11 +268,13 @@ class Matrix(ApiRequest):
             timeout (int): requests module timeout used in ApiRequest.query
                 method
             debug (bool): enable/disable debugging in requests module
+            verify(bool): SSL verification is turned on by default
+                and can be turned off using this method.
         """
         super().__init__(
             log, user, token,
             base_url, matrix_path,
-            timeout, debug
+            timeout, debug, verify
         )
         self.user = user
 
@@ -360,7 +372,8 @@ class SynapseAdmin(ApiRequest):
         ApiRequest (object): parent class containing general properties and
             methods for requesting REST API's
     """
-    def __init__(self, log, user, token, base_url, admin_path, timeout, debug):
+    def __init__(self, log, user, token, base_url, admin_path, timeout, debug,
+                 verify):
         """Initialize the SynapseAdmin object
 
         Args:
@@ -375,11 +388,13 @@ class SynapseAdmin(ApiRequest):
             timeout (int): Requests module timeout used in ApiRequest.query
                 method
             debug (bool): enable/disable debugging in requests module
+            verify(bool): SSL verification is turned on by default
+                and can be turned off using this method.
         """
         super().__init__(
             log, user, token,
             base_url, admin_path,
-            timeout, debug
+            timeout, debug, verify
         )
         self.user = user
 
--- a/synadm/cli/__init__.py
+++ b/synadm/cli/__init__.py
@@ -386,7 +386,7 @@ def root(ctx, verbose, batch, output, config_file):
     method set by --server-discovery."""
 )
 @click.option(
-    "--ssl_verify", "-n",  default=True, show_default=True,
+    "--ssl-verify", "-i",  is_flag=True, default=True, show_default=True,
     help="""Whether or not SSL certificates should be verified. Set to False
     to allow self-signed certifcates."""
 )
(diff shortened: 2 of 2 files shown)


JOJ0 (changes requested):
Hi @heurtematte, thanks for the submission, I'm sure this is a very useful features for a lot of other synadm users as well! Nice!

One thing that comes up in a first and unthorough review is that you define boolean as for the new verify flag on cli level but at some point in the code it is changed / has a default value of None which not really is boolean. Is there a reason for that or could we change it to stay True/False verywhere?

Also, some tiny formatting issues are detected by our linter (flake8) in the CI.

Thanks again and speak soon!

heurtematte:
> Is there a reason for that or could we change it to stay True/False everywhere?

I wanted to keep the method `query` in the same state, and offer the possibility to override the value programmatically if really necessary over configuration setting, and testing None is easire than a boolean value.   

```
def query(self, method, urlpart, params=None, data=None, token=None,
              base_url_override=None, verify=None):
```

Tell me if you want me to change this behavior.

I'll check the linter.

synadm/api.py
```diff
@@ -55,6 +56,8 @@ def __init__(self, log, user, token, base_url, path, timeout, debug):
                 base_url to form the basis for all API endpoint paths
             timeout (int): requests module timeout used in query method
             debug (bool): enable/disable debugging in requests module
+            verify(bool): SSL verification is turned on by default
+                and can be turned off using this method.
```
JOJ0, line 60: I'd prefer the wording: ....using this argument.

Also, is the type now still bool?

JOJ0 (changes requested):
Sorry for being picky but we are chabging somethin very fundamental here and to make sure synadm behaves correctly n all use cases. thanks!

synadm/cli/__init__.py
```diff
@@ -376,10 +380,15 @@ def root(ctx, verbose, batch, output, config_file):
     default value 'auto-retrieval' will try to discover the name using the
     method set by --server-discovery."""
 )
+@click.option(
+    "--ssl_verify", "-n",  default=True,
+    help="""Verify certificate (Default True, False allow to manage \
+        selfsigned certificate)."""
+)
```
JOJ0, line 387: Default true can be specified by Click's show_default=True flag. Use that and I think we could then rephares this help text to something like: "Whether or not SSL certificates should be verified. Set to False to allow self-signed certifcates." or maybe even shorter because ssl_verify tells it all to a regular Matrix user/admin already:

Set to False to allow self-signed certifcates.

JOJ0:
Hi @heurtematte can I assist with anything else?

There are some things to fix. No hurry in doing that right away. Take your time, the merge of this new feature can wait. I just would like to know if you're still on it.

In the meantime I did some testing and another thing comes up. It seems the actual value set in the config is not displayed in the interactive configurator. This is misleading. Have a look how other config settings do it. Any non-sensitive information should be printed while the user reconfigures.

```
(synadm38) ~/git/heurtematte_synadm (master) $ grep verify ~/.config/synadm.yaml 
ssl_verify: 'false'
(synadm38) ~/git/heurtematte_synadm (master) $ 
(synadm38) ~/git/heurtematte_synadm (master) $ 
(synadm38) ~/git/heurtematte_synadm (master) $ 
(synadm38) ~/git/heurtematte_synadm (master) $ synadm config 
Running configurator...
Synapse admin user name [admin]: 
Synapse admin user token [REDACTED]: 
Synapse base URL [https://example.org]: 
Synapse admin API path [/_synapse/admin]: 
Matrix API path [/_matrix]: 
Default output format (yaml, json, human, pprint) [yaml]: 
Default http timeout [3600]: 
Homeserver name (auto-retrieval or matrix.DOMAIN) [auto-retrieval]: 
Verify certificate (Default True, False allow to manage                 selfsigned certificate) [True]: 
```

heurtematte:
sorry a bit busy! I'll take a look by the end of the week.

synadm/cli/__init__.py
```diff
@@ -451,6 +462,11 @@ def get_redacted_token_prompt(cli_token):
             "Homeserver name (auto-retrieval or matrix.DOMAIN)",
             default=homeserver if homeserver else helper.config.get(
                 "homeserver", homeserver)),
+        "ssl_verify": click.prompt(
+            "Verify certificate (Default True, False allow to manage \
+                selfsigned certificate)",
```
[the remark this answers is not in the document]
Ascurius: Please note that we have currently not defined the default values of the other config variables explicitly in the help string. Because if you run the configurator the current value (by default is set to the values defined in `APIHelper.CONFIG`) is presented behind each prompt within brackets, so the user exactly knows which value the corresponding config variable is currently set to. For example it looks like `Synapse base URL [http://localhost:8008]`. 
Therefore I don't think that there is a need to explicitly state the default value within the help string

synadm/cli/__init__.py
heurtematte: Yaml seems to support different case: |true|True|TRUE|false|False|FALSE
According to https://yaml.org/type/bool.html

synadm/api.py
heurtematte: Good question! I'll check

synadm/cli/__init__.py
```diff
@@ -129,8 +130,8 @@ def load(self):
         except Exception as error:
             self.log.error("%s while reading configuration file", error)
         for key, value in self.config.items():
-            if not value:
-                self.log.error("Config entry missing: %s", key)
+            if value is None:
+                self.log.error("Config entry missing: %s, %s", key, value)
```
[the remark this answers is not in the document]
heurtematte: Good catch. I roll back my modification.

synadm/api.py
```diff
@@ -108,10 +112,14 @@ def query(self, method, urlpart, params=None, data=None, token=None,
             self.log.debug("Token override! Adjusting headers.")
             self.headers["Authorization"] = "Bearer " + token
 
+        override_verify = self.verify
+        if verify is not None:
+            override_verify = verify
+
```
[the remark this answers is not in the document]
heurtematte: exactly!

synadm/api.py
```diff
@@ -375,11 +388,13 @@ def __init__(self, log, user, token, base_url, admin_path, timeout, debug):
             timeout (int): Requests module timeout used in ApiRequest.query
                 method
             debug (bool): enable/disable debugging in requests module
+            verify(bool): SSL verification is turned on by default
+                and can be turned off using this method.
```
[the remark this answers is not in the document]
heurtematte: ok

synadm/cli/__init__.py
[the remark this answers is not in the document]
heurtematte: sounds good!

synadm/cli/__init__.py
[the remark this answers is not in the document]
heurtematte: I remember why now. 

If I set false to the ssl_verify configuration. I have this message

```
ERROR Config entry missing: ssl_verify, False
```

So I must deal with string value and boolean value.

synadm/api.py
heurtematte: We can keep bool type

synadm/cli/__init__.py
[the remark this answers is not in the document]
JOJ0: I understand!

JOJ0:
Hi @heurtematte, thanks for your changes last week. We've been busy with other things in the project but I think around next week might be able to conintue work here. We'd like to do some testing and tiny changes to finalize this. Do you mind if we take over your branch to incorporate said finalization? Thanks!

JOJ0:
Hi @JacksonChen666, I've added a couple of formatting but also functional fixes to make a real boolean value in the config work.

Please could you possibly try this in your testing environment:

- What happens if the config is missing at all? For example if a user upgrades. -> Should be just ignored and the default of "True" is used.
- What happens if the user issues `synadm config` for the first time after an upgrade (withouth the option)
- What happens if rubbish is written into the config? E.g `ssl_verify: bollocks text`

Thanks a lot!

synadm/cli/__init__.py
```diff
@@ -376,10 +385,15 @@ def root(ctx, verbose, batch, output, config_file):
...
+    "--ssl-verify", "-i",  is_flag=True, default=True, show_default=True,
+    help="""Whether or not SSL certificates should be verified. Set to False
+    to allow self-signed certifcates."""
+)
 @click.pass_obj
 def config_cmd(helper, user_, token, base_url, admin_path, matrix_path,
-               output, timeout, server_discovery, homeserver):
-    """ Modify synadm's configuration.
+               output, timeout, server_discovery, homeserver, ssl_verify):
+    """ Modify test synadm's configuration.
```
JacksonChen666, line 396: test?

synadm/cli/__init__.py
```diff
@@ -376,10 +385,15 @@ def root(ctx, verbose, batch, output, config_file):
     default value 'auto-retrieval' will try to discover the name using the
     method set by --server-discovery."""
 )
+@click.option(
+    "--ssl-verify", "-i",  is_flag=True, default=True, show_default=True,
```
JacksonChen666, line 389: `default=True` is odd in this case, as defaults aren't set in any options above but instead below in the `click.prompt` things.

this makes for a UI inconsistency where all other configs use the default value of whatever is set (in `synadm config`) in the options except for `--ssl-verify`.

JacksonChen666 (changes requested):
> Please could you possibly try this in your testing environment:
>     * What happens if rubbish is written into the config? E.g `ssl_verify: bollocks text`

default value is used **regardless of config values**, even if configured with `False` in `synadm config` it will default to `True` (very likely not intended behavior).

I have made a comment in the place where I think the issue is, plus something else I noticed.

synadm/cli/__init__.py
JOJ0: Oh, that seems to be a leftover of some debugging by the OP or something. Corrected.

synadm/cli/__init__.py
JOJ0: Yes! You are absolutely right, we don't handle defaults that way! Defaults should always be handled via APIHelper.CONFIG and the click.prompt things as you noticed!  Good catch! I fixed this. 

I think now the behaviour is quite fine. Can you please give it another testrun?

To me it looks like this: 
- We silently default to true if the option is missing.
- We display an error when the option is manually set wrongly to some non-boolean string.
  but we don't quit synadm but continue with subsequent errors. This is good, since otherwhise the configurator could not be run. For example:
  
 ```
 (synadm38) ~/git/heurtematte_synadm (master) $ synadm -vv user details testuser2
DEBUG Config entry read. user: admin
DEBUG Config entry read. token: REDACTED
DEBUG Config entry read. base_url: https://peek-a-boo.at
DEBUG Config entry read. admin_path: /_synapse/admin
DEBUG Config entry read. matrix_path: /_matrix
DEBUG Config entry read. timeout: 3600
DEBUG Config entry read. server_discovery: well-known
DEBUG Config entry read. homeserver: peek-a-boo.at
ERROR Config value error: ssl_verify, crap must be boolean
DEBUG Config entry read. ssl_verify: crap
DEBUG Config entry read. format: yaml
DEBUG Formatter in use: yaml - <function dump at 0x7f69b9e70e18>
DEBUG A proper localpart was passed, generating MXID for local homeserver.
INFO  Querying get on https://peek-a-boo.at/_synapse/admin/v2/users/@testuser2:peek-a-boo.at
ERROR OSError while querying Synapse: Could not find a suitable TLS CA certificate bundle, invalid path: crap
User details could not be fetched.
(synadm38) ~/git/heurtematte_synadm (master) $ 
```

PR #90, opened 2023-02-03
Make it possible to change user type

gergelypolonkai:
The user_type value is not checked (yet?) the code, although the spec allows only types `bot` and `support` (and the API will respond with `M_UNKNOWN` if an invalid value is sent). If it should be checked, let me know and i will update the PR.

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -557,7 +557,8 @@ class SynapseAdmin(ApiRequest):
         if deactivation == "activate":
             data.update({"deactivated": False})
         if user_type:
-            data.update({"user_type": None if 'null' else user_type})
+            data.update({"user_type": None if user_type == 'null' else
+                         user_type})
         return self.query("put", f"v2/users/{user_id}", data=data)
 
     def user_whois(self, user_id):
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -556,7 +556,8 @@ class SynapseAdmin(ApiRequest):
             data.update({"deactivated": True})
         if deactivation == "activate":
             data.update({"deactivated": False})
-        data.update({"user_type": user_type})
+        if user_type:
+            data.update({"user_type": user_type})
         return self.query("put", f"v2/users/{user_id}", data=data)
 
     def user_whois(self, user_id):
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -557,7 +557,10 @@ class SynapseAdmin(ApiRequest):
         if deactivation == "activate":
             data.update({"deactivated": False})
         if user_type:
-            data.update({"user_type": user_type})
+            if user_type == 'null':
+                data.update({"user_type": None})
+            else:
+                data.update({"user_type": user_type})
         return self.query("put", f"v2/users/{user_id}", data=data)
 
     def user_whois(self, user_id):
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -557,10 +557,7 @@ class SynapseAdmin(ApiRequest):
         if deactivation == "activate":
             data.update({"deactivated": False})
         if user_type:
-            if user_type == 'null':
-                data.update({"user_type": None})
-            else:
-                data.update({"user_type": user_type})
+            data.update({"user_type": None if 'null' else user_type})
         return self.query("put", f"v2/users/{user_id}", data=data)
 
     def user_whois(self, user_id):
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -534,7 +534,7 @@ class SynapseAdmin(ApiRequest):
         return self.query("post", f"v1/users/{user_id}/login", data=data)
 
     def user_modify(self, user_id, password, display_name, threepid,
-                    avatar_url, admin, deactivation):
+                    avatar_url, admin, deactivation, user_type):
         """ Create or update information about a given user
 
         Threepid should be passed as a tuple in a tuple
@@ -556,6 +556,7 @@ class SynapseAdmin(ApiRequest):
             data.update({"deactivated": True})
         if deactivation == "activate":
             data.update({"deactivated": False})
+        data.update({"user_type": user_type})
         return self.query("put", f"v2/users/{user_id}", data=data)
 
     def user_whois(self, user_id):
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -358,10 +358,18 @@ class UserModifyOptionGroup(RequiredAnyOptionGroup):
     removes their active access tokens, resets their password, kicks them out
     of all rooms and deletes third-party identifiers (to prevent the user
     requesting a password reset). See also "user deactivate" command.""")
+@optgroup.option(
+    "--user-type", "user_type", type=str,
+    help="""Change the type of the user. Currently allowed values are 'bot'
+    and support.""")
+@optgroup.option(
+    "--clear-user-type", "clear_user_type", is_flag=True,
+    help="""Clear the user_type field of the user.""")
 @click.pass_obj
 @click.pass_context
 def modify(ctx, helper, user_id, password, password_prompt, display_name,
-           threepid, avatar_url, admin, deactivation):
+           threepid, avatar_url, admin, deactivation, user_type,
+           clear_user_type):
     """ Create or modify a local user. Provide matrix user ID (@user:server)
     as argument.
     """
@@ -372,6 +380,10 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
     if deactivation == "deactivate" and (password_prompt or password):
         click.echo(
             "Deactivating a user and setting a password doesn't make sense.")
+        raise SystemExit(1)
+
+    if clear_user_type and user_type:
+        click.echo("Use either --user-type or --clear-user-type, not both.")
         raise SystemExit(1)
 
     mxid = helper.generate_mxid(user_id)
@@ -390,6 +402,9 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
                             f"{t_key} is probably not a supported medium "
                             "type. Threepid medium types according to the "
                             "current matrix spec are: email, msisdn.")
+        elif key == "clear_user_type":
+            if value:
+                click.echo("user_type: null")
         elif value not in [None, {}, []]:  # only show non-empty (aka changed)
             click.echo(f"{key}: {value}")
 
@@ -411,7 +426,8 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
     if sure:
         modified = helper.api.user_modify(
             mxid, password, display_name, threepid,
-            avatar_url, admin, deactivation)
+            avatar_url, admin, deactivation,
+            None if clear_user_type else user_type)
         if modified is None:
             click.echo("User could not be modified.")
             raise SystemExit(1)
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -359,17 +359,17 @@ class UserModifyOptionGroup(RequiredAnyOptionGroup):
     of all rooms and deletes third-party identifiers (to prevent the user
     requesting a password reset). See also "user deactivate" command.""")
 @optgroup.option(
-    "--user-type", "user_type", type=str,
-    help="""Change the type of the user. Currently allowed values are 'bot'
-    and 'support'.""")
-@optgroup.option(
-    "--clear-user-type", "clear_user_type", is_flag=True,
-    help="""Clear the user_type field of the user.""")
+    "--user-type", type=str, default=None, show_default=True,
+    help="""Change the type of the user. Currently understood by the Admin API
+    are 'bot' and 'support'. Use 'regular' to create a regular Matrix user
+    (which effectively sets the user-type to 'null'). If the --user-type option
+    is omitted when modifying an existing user, the user-type will not be
+    manipulated. If the --user-type option is omitted when creating a new user,
+    a regular user will be created.""")
 @click.pass_obj
 @click.pass_context
 def modify(ctx, helper, user_id, password, password_prompt, display_name,
-           threepid, avatar_url, admin, deactivation, user_type,
-           clear_user_type):
+           threepid, avatar_url, admin, deactivation, user_type):
     """ Create or modify a local user. Provide matrix user ID (@user:server)
     as argument.
     """
@@ -380,10 +380,6 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
     if deactivation == "deactivate" and (password_prompt or password):
         click.echo(
             "Deactivating a user and setting a password doesn't make sense.")
-        raise SystemExit(1)
-
-    if clear_user_type and user_type:
-        click.echo("Use either --user-type or --clear-user-type, not both.")
         raise SystemExit(1)
 
     mxid = helper.generate_mxid(user_id)
@@ -402,9 +398,8 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
                             f"{t_key} is probably not a supported medium "
                             "type. Threepid medium types according to the "
                             "current matrix spec are: email, msisdn.")
-        elif key == "clear_user_type":
-            if value:
-                click.echo("user_type: null")
+        elif key == "user_type" and value == 'regular':
+            click.echo("user_type: null")
         elif value not in [None, {}, []]:  # only show non-empty (aka changed)
             click.echo(f"{key}: {value}")
 
@@ -427,7 +422,7 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
         modified = helper.api.user_modify(
             mxid, password, display_name, threepid,
             avatar_url, admin, deactivation,
-            None if clear_user_type else user_type)
+            None if user_type == 'regular' else user_type)
         if modified is None:
             click.echo("User could not be modified.")
             raise SystemExit(1)
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -361,7 +361,7 @@ class UserModifyOptionGroup(RequiredAnyOptionGroup):
 @optgroup.option(
     "--user-type", "user_type", type=str,
     help="""Change the type of the user. Currently allowed values are 'bot'
-    and support.""")
+    and 'support'.""")
 @optgroup.option(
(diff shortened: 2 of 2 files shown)


synadm/cli/user.py
```diff
@@ -374,6 +382,10 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
             "Deactivating a user and setting a password doesn't make sense.")
         raise SystemExit(1)
 
+    if clear_user_type and user_type:
+        click.echo("Use either --user-type or --clear-user-type, not both.")
+        raise SystemExit(1)
+
```
JOJ0, line 388: Instead of this sanity check, please use the mutual exclusive option group. Look at the imports of cli/modules to find its exact name.

synadm/api.py
```diff
@@ -556,6 +556,7 @@ def user_modify(self, user_id, password, display_name, threepid,
             data.update({"deactivated": True})
         if deactivation == "activate":
             data.update({"deactivated": False})
+        data.update({"user_type": user_type})
```
JOJ0, line 559: What if it's None? What does the API respond then? Valid? Was that the M_UNKNOWN response you were referring to?

synadm/cli/user.py
```diff
@@ -411,7 +426,8 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
     if sure:
         modified = helper.api.user_modify(
             mxid, password, display_name, threepid,
-            avatar_url, admin, deactivation)
+            avatar_url, admin, deactivation,
+            None if clear_user_type else user_type)
```
JOJ0, line 430: Ah ok, here you make sure user_type is always passed. Good!

JOJ0 (changes requested):
Regarding your questionaround sanity check: In the end we'd like a user to not having to specify --user-type, it should have a resonable default. Post some example cli calls, that would help me review without testing myself :-) Cheers!

synadm/api.py
gergelypolonkai: Yes, `user_type` can be `null` in the JSON in which case the field gets cleared (which is basically the “default” option).

gergelypolonkai:
> we'd like a user to not having to specify --user-type

You don’t have to, but you can. Not having a user-type set (which can be achieved by using the `--clear-user-type` command line option) is the default. According to the spec, a user can have their user type set to `null` (default), `"bot"`, and `"support"` (whatever that may mean). Specifying any other type on the API call will result in `M_UNKNOWN`.

synadm/cli/user.py
gergelypolonkai: If i add them to a `MutuallyExclusiveOptionGroup` and i only pass the `--user-type` option to the `user modify` command, it tells me that i have to pass one of the other options (as if the mutually exclusive group is not part of the `RequireAnyOptionGroup`).

gergelypolonkai: Ah… from the [documentation](https://click-option-group.readthedocs.io/en/latest/tutorial.html#limitations) of click-option-group:

> The package does not support nested option groups (option subgroups). This is intentional. Nested option groups complicate the implementation, API and CLI and most often it is not necessary.

So it seems either the current implementation remains, or the user type change can only be done together with changing anything else.

JOJ0: Ah ok then this is the reason why the password thing right above is done with a conditian as well. Ok then please leave as as. Not worth the hassle! :-)

synadm/api.py
JOJ0: Basically that's ok but the bad thing is that if any other property of a user gets changed, the user_type is always cleared. Do we have this behavior somewhere else already by any chance?

gergelypolonkai: Now that you mention it, it seems the activate/deactivate thing works just like that 😁 iʼll check tomorrow as itʼs bedtime for me now.

gergelypolonkai: I was right; the activate/deactivate feature works just like this. If i don’t add `--user-type` nor `--clear-user-type` to the command line, it won’t get updated:

```
$ synadm user modify --password abcdef @test:matrix.local
Current user account settings:
{"name": "@test:matrix.local "is_guest": 0, "admin": false, "consent_version": null, "consent_ts": null, "consent_server_notice_sent": null, "appservice_id": null, "creation_ts": 1457796444, "user_type": null, "deactivated": false, "shadow_banned": false, "displayname": "Test Account", "avatar_url": null, "threepids": [], "external_ids": [], "erased": false}
User account settings to be modified:
Password will be set as provided on command line.
Are you sure you want to modify user? (y/N):
```

JOJ0: Just read the api docs: https://matrix-org.github.io/synapse/develop/admin_api/user_admin_api.html#create-or-modify-account

user_type is optional, I think you just need to make sure it isnt included at all when both of the 2 new options are not given on cli. So nothing gets cleared 🎉 Currently you always include it in the data sent to the endpoint.

JOJ0: @gergelypolonkai I've just added the missing condition as suggested in my previous post. I've just edited your feature branch. I hope that's ok.

JOJ0: This change was necessary, otherwise this would happen as with your original code:
```
(synadm38) ~/git/gergelypolonkai_synadm (user-type-edit) $ synadm user modify --display-name "TESTUSER1" testuser1
Current user account settings:
admin: 0
appservice_id: null
avatar_url: null
consent_server_notice_sent: null
consent_version: null
creation_ts: 1573653091
deactivated: 0
displayname: Test User 1
external_ids: []
is_guest: 0
name: '@testuser1:peek-a-boo.at'
password_hash: $secret
shadow_banned: null
threepids: []
user_type: bot

User account settings to be modified:
display_name: TESTUSER1
Are you sure you want to modify user? (y/N): y
admin: 0
appservice_id: null
avatar_url: null
consent_server_notice_sent: null
consent_version: null
creation_ts: 1573653091
deactivated: 0
displayname: TESTUSER1
external_ids: []
is_guest: 0
name: '@testuser1:peek-a-boo.at'
password_hash: $secret
shadow_banned: null
threepids: []
user_type: null
```

JOJ0 (approved):

JOJ0:
We are almos ready to merge. I see a tiny thing in --help: Allowed values are formatted differently. 'bot' has single quotes and 'support' has it. Please add that.

Other than LGTM. Final thoughts from @Ascurius and/or @JacksonChen666 appreciated! One usability thing I'm wondering about is: Is `--clear-user-type` really necessary _or_ would something like `--user-type clear` or `user-type null` be a good alternative. I'm really not sure about whether that would be a "clean" and expected solution for a unix-shell-util. Give me your thoughts! Thanks!

synadm/cli/user.py
```diff
@@ -358,10 +358,18 @@ def name_extra(self):
     removes their active access tokens, resets their password, kicks them out
     of all rooms and deletes third-party identifiers (to prevent the user
     requesting a password reset). See also "user deactivate" command.""")
+@optgroup.option(
+    "--user-type", "user_type", type=str,
+    help="""Change the type of the user. Currently allowed values are 'bot'
+    and support.""")
```
JacksonChen666, line 364: Inconsistent use of single quotes, here's what I suggest:
```suggestion
    and 'support'.""")
```

JacksonChen666 (changes requested):

JacksonChen666:
> One usability thing I'm wondering about is: Is `--clear-user-type` really necessary _or_ would something like `--user-type clear` or `user-type null` be a good alternative? I'm really not sure about whether that would be a "clean" and expected solution for a unix-shell-util. Give me your thoughts! Thanks!

I'm not sure either... `--user-type clear` could make sense but `--clear-user-type` also makes sense but I don't know which I would prefer (maybe the former?). `--user-type null` is quite specific and doesn't really say what it is, so maybe not that.

JOJ0:
Hi @gergelypolonkai, I would like to take over this PR entirely, since we would like to include it in the next synadm release and there is some things left to do. I hope that is ok for you. Thanks for the submission of a useful feature!

JOJ0:
> > One usability thing I'm wondering about is: Is `--clear-user-type` really necessary _or_ would something like `--user-type clear` or `user-type null` be a good alternative? I'm really not sure about whether that would be a "clean" and expected solution for a unix-shell-util. Give me your thoughts! Thanks!
> 
> I'm not sure either... `--user-type clear` could make sense but `--clear-user-type` also makes sense but I don't know which I would prefer (maybe the former?). `--user-type null` is quite specific and doesn't really say what it is, so maybe not that.

I think I have made up my mind what seems to feel natural to me:

- Remove --clear-user-type
- Document --user-type well so it explains that 3 options are currently valid (or even just prevent any other options by Click's choice feature?)
  - bot
  - support
  - default (This is technically setting to 'null', thus a regular user, actually a default matrix user, nothing else)
  
  @JacksonChen666 @Ascurius what do you think?

The term default is not stated as such in the admin api docs (> One usability thing I'm wondering about is: Is `--clear-user-type` really necessary _or_ would something like `--user-type clear` or `user-type null` be a good alternative? I'm really not sure about whether that would be a "clean" and expected solution for a unix-shell-util. Give me your thoughts! Thanks!

I'm not sure either... `--user-type clear` could make sense but `--clear-user-type` also makes sense but I don't know which I would prefer (maybe the former?). `--user-type null` is quite specific and doesn't really say what it is, so maybe not that.

I think I have made up my mind what seems to feel natural to me:

Why not make this ) but I still think that it's a very logical and self-explaining thing if we just call this user_type default

JOJ0:
> I think I have made up my mind what seems to feel natural to me:
> 
> * Remove `--clear-user-type`
> * Document `--user-type` well so it explains that 3 options are currently valid (or even just prevent any other options by Click's choice feature?)
>   
>   * bot
>   * support
>   * default (This is technically setting to 'null', thus a regular user, actually a default matrix user, nothing else)

I just wanted to put these changes into "suggested changes" with this neat feature here but I decided not to. It's quite some changes necessary and it's easier and less errror-prone if I just push that to the branch as one commit. I'll wait with that until you tell me your opinions about this idea.

Thanks in advance!

JacksonChen666 (approved):
Looks good to me. There are a few things I noticed:
- `--clear-user-type` or `--user-type null`, I'm not sure.
- `--user-type` will accept any argument, including non-existing user types. Should that be restricted to known valid options?

JOJ0:
> 

I really don't like `--user-type null`, it is not self-explanatory at all. If there is no strong opinions against `--user-type default`, I will use that one instead. In my opinion it tells everything to a user who did not read the API docs at all. A standard user / a default user will be created / existing user will become that. 

I was not sure about the restriction of the possible arguments either but would now say that it's better we leave this to unrestricted for now, since it could happen that new things are added to the list. The API will give us a nice response of M_UKNOWN

JOJ0:
> Looks good to me. There are a few things I noticed:
> 
> * `--clear-user-type` or `--user-type null`, I'm not sure.

I really don't like `--user-type null`, it is not self-explanatory at all. If there is no strong opinions against `--user-type default`, I will use that one instead. In my opinion it tells everything to a user who did not read the API docs at all. A standard user / a default user will be created / existing user will become that. 

> * `--user-type` will accept any argument, including non-existing user types. Should that be restricted to known valid options?

I was not sure about the restriction of the possible arguments either but would now say that it's better we leave this to unrestricted for now, since it could happen that new things are added to the list. The API will give us a nice response of M_UKNOWN

JacksonChen666:
> I really don't like `--user-type null`, it is not self-explanatory at all. If there is no strong opinions against `--user-type default`, I will use that one instead. In my opinion it tells everything to a user who did not read the API docs at all. A standard user / a default user will be created / existing user will become that.

I would lean towards a `default` user, Though I'm not sure what a user would think about a "default" user cause that doesn't seem extremely obvious either...

`--clear-user-type` is kind of awkward. What about `--user-type clear`?

synadm/cli/user.py
```diff
@@ -359,17 +359,17 @@ def name_extra(self):
...
-    and 'support'.""")
-@optgroup.option(
-    "--clear-user-type", "clear_user_type", is_flag=True,
-    help="""Clear the user_type field of the user.""")
+    "--user-type", type=str, default=None, show_default=True,
+    help="""Change the type of the user. Currently understood by the Admin API
+    are 'bot' and 'support'. Use 'regular' to create a regular Matrix user
+    (which effectively sets the user-type to 'null'). If the --user-type option
+    is omitted when modifying an existing user, the user-type will not be
+    manipulated. If the --user-type option is omitted when creating a new user,
```
JacksonChen666, line 367: > If the --user-type option is omitted when creating a new user

But this command is about modifying, not creating, right?

JacksonChen666 (changes requested):
```
> synadm user modify '@130:localhost:8480' --user-type support
Current user account settings:
{"name": "@130:localhost:8480", "is_guest": 1, "admin": true, "consent_version": null, "consent_ts": null, "consent_server_notice_sent": null, "appservice_id": null, "creation_ts": 1676753770, "user_type": "bot", "deactivated": false, "shadow_banned": false, "displayname": "130", "avatar_url": null, "threepids": [], "external_ids": [], "erased": false}
User account settings to be modified:
user_type: support
Are you sure you want to modify user? (y/N): y
{"name": "@130:localhost:8480", "is_guest": 1, "admin": true, "consent_version": null, "consent_ts": null, "consent_server_notice_sent": null, "appservice_id": null, "creation_ts": 1676753770, "user_type": "support", "deactivated": false, "shadow_banned": false, "displayname": "130", "avatar_url": null, "threepids": [], "external_ids": [], "erased": false}

> synadm user modify '@130:localhost:8480' --user-type regular
Current user account settings:
{"name": "@130:localhost:8480", "is_guest": 1, "admin": true, "consent_version": null, "consent_ts": null, "consent_server_notice_sent": null, "appservice_id": null, "creation_ts": 1676753770, "user_type": "support", "deactivated": false, "shadow_banned": false, "displayname": "130", "avatar_url": null, "threepids": [], "external_ids": [], "erased": false}
User account settings to be modified:
user_type: null
Are you sure you want to modify user? (y/N): y
{"name": "@130:localhost:8480", "is_guest": 1, "admin": true, "consent_version": null, "consent_ts": null, "consent_server_notice_sent": null, "appservice_id": null, "creation_ts": 1676753770, "user_type": "support", "deactivated": false, "shadow_banned": false, "displayname": "130", "avatar_url": null, "threepids": [], "external_ids": [], "erased": false}
```

**This is broken** and I don't know why (or it doesn't work (or not for guest accounts)). Also applies to revoking admin privileges (unrelated to this PR).

Also, the "regular" option feels kinda odd.

JOJ0:
I had just realized this issue but needed a break...consider my latest commit a draft

synadm/cli/user.py
JOJ0: Nope, it's both! See admin api docs, there is one api for both creating and modifying.

Yeah an alias name for `modify` to `create` would be great! I want that since ages...

synadm/api.py
```diff
@@ -557,7 +557,10 @@ def user_modify(self, user_id, password, display_name, threepid,
         if deactivation == "activate":
             data.update({"deactivated": False})
         if user_type:
-            data.update({"user_type": user_type})
+            if user_type == 'null':
```
JacksonChen666, line 560: could be instead:
```python
data.update({"user_type": None if user_type == 'null' else user_type})
```

JOJ0:
Ok, setting user-type now works. I worked around the "poor desing" of user_modify() by just passing a string 'null'.
Well, yes, that is ugly, but it will work. Rerfactoring of that method should be tackled in another PR, I agree. Thanks for opening the issue.

So the only thing that remains that you still think we don't have a good name for creating a regular, a normal, a default matrix user-type. I am running out of ideas and still think putting something like 'null' or 'clear' is not good usability, nor is ist it obvious what the user should expect. They would need to read the admin api docs to understand what that means.

So that's why I still would like to ask you for a suggestion of how you think from a user-perspecitive what should that option be called. Consider these english sentences, which one does make the most self-explanatory sense to you?

- A user of the type 'clear'
- A user of the type 'default'
- A user of the type 'regular'
- A user of the type 'null

synadm/api.py
JOJ0: yeah that's nicer. I like oneliners like that but will check if the linelength is sufficient......

JOJ0: Oh, well that fits easily. Changed it. Thanks for the suggestion. Much better well....much less ugly now ;-))

JOJ0:
Ok so besides the final decision for option name I think this is good to go now.

JOJ0:
One more thing. We should consider how the help looks like for a final decision. I tried to put all the behaviour possibilities into it. Hope it's easily understood:

```
    --user-type TEXT              Change the type of the user. Currently
                                  understood by the Admin API are 'bot' and
                                  'support'. Use 'regular' to create a regular
                                  Matrix user (which effectively sets the
                                  user-type to 'null'). If the --user-type
                                  option is omitted when modifying an existing
                                  user, the user-type will not be manipulated.
                                  If the --user-type option is omitted when
                                  creating a new user, a regular user will be
                                  created.
```

synadm/api.py
```diff
@@ -556,6 +556,8 @@ def user_modify(self, user_id, password, display_name, threepid,
             data.update({"deactivated": True})
         if deactivation == "activate":
             data.update({"deactivated": False})
+        if user_type:
+            data.update({"user_type": None if 'null' else user_type})
```
JacksonChen666, line 560:
```suggestion
            data.update({"user_type": None if user_type == 'null' else
                         user_type})
```

`'null'` always evaluates to true, so user type will always be reset regardless of passed value.

JacksonChen666 (changes requested):
@JOJ0 `synadm/api.py` Line 560 is incorrect. Check my suggestion that fixes, or pull from my branch (`git pull https://github.com/JacksonChen666/synadm user-type-edit`).

(And no, I didn't mess my suggestion up, I checked.)

After applying my change correctly (either using `git pull https://github.com/JacksonChen666/synadm user-type-edit` or using GitHub to commit the suggestion), you can assume an approval from me.

JOJ0:
Whoops, thanks for spotting this embarrassing mistake! :-)

JacksonChen666 (approved):
(this is a blind-ish approval, but I have already reviewed with my fix and my fix has been applied, so it's not so blind?)

PR #92, opened 2023-02-17
Finalize PR #84 and improve --output helptext

JOJ0:
- Keep -o json as the new default as the PR suggests.
- Add a new mode -o jsonmini that still outputs oneline-json as the original json output mode did.
- Improve handling of help text. We didn't have any description of the modes with the global --output flag. It was hidden with the config -o subcommand. It's shown in both places now.

### DIFF
--- a/synadm/cli/__init__.py
+++ b/synadm/cli/__init__.py
@@ -33,13 +33,13 @@ import re
 from synadm import api
 
 
-output_format_help = """'human' gives a tabular or list view depending on the
-fetched data. This mode needs your terminal to be quite wide! 'json' returns
-formatted json, 'minified' is minified json and displays exactly as the API
-responded. 'pprint' shows a formatted output with the help of Python's built-in
-pprint module. 'yaml' is a compromise between human- and machine-readable
-output, it doesn't need as much terminal width as 'human' does and is the
-default on fresh installations."""
+output_format_help = """The 'human' mode gives a tabular or list view depending
+on the fetched data, but often needs a lot of horizontal space to display
+correctly. 'json' returns formatted json. 'minified' is minified json, suitable
+for scripting purposes. 'pprint' shows a formatted output with the help of
+Python's built-in pprint module. 'yaml' is a compromise between human- and
+machine-readable output, it doesn't need as much terminal width as 'human' does
+and is the default on fresh installations."""
 
 
 def humanize(data):
--- a/synadm/cli/__init__.py
+++ b/synadm/cli/__init__.py
@@ -358,7 +358,8 @@ def root(ctx, verbose, batch, output, config_file):
     help="""The time in seconds synadm should wait for responses from admin
     API's or Matrix API's. The default is 7 seconds. """)
 @click.option(
-    "--output", "-o", type=click.Choice(["yaml", "json", "human", "pprint"]),
+    "--output", "-o", type=click.Choice([
+        "yaml", "json", "minified", "human", "pprint"]),
     help=f"""How synadm displays data by default. {output_format_help} The
     default output format can always be overridden by using the global
     --output/-o switch (eg 'synadm -o pprint user list').""")
@@ -452,7 +453,8 @@ def config_cmd(helper, user_, token, base_url, admin_path, matrix_path,
         "format": click.prompt(
             "Default output format",
             default=output if output else helper.config.get("format", output),
-            type=click.Choice(["yaml", "json", "jsonmini", "human", "pprint"])),
+            type=click.Choice([
+                "yaml", "json", "minified", "human", "pprint"])),
         "timeout": click.prompt(
             "Default http timeout",
             default=timeout if timeout else helper.config.get(
--- a/synadm/cli/__init__.py
+++ b/synadm/cli/__init__.py
@@ -33,6 +33,15 @@ import re
 from synadm import api
 
 
+output_format_help = """'human' gives a tabular or list view depending on the
+fetched data. This mode needs your terminal to be quite wide! 'json' returns
+formatted json, 'jsonmini' is minified json and displays exactly as the API
+responded. 'pprint' shows a formatted output with the help of Python's built-in
+pprint module. 'yaml' is a compromise between human- and machine-readable
+output, it doesn't need as much terminal width as 'human' does and is the
+default on fresh installations."""
+
+
 def humanize(data):
     """ Try to display data in a human-readable form:
     - Lists of dicts are displayed as tables.
@@ -62,6 +71,7 @@ class APIHelper:
     FORMATTERS = {
         "pprint": pprint.pformat,
         "json": json_pretty,
+        "jsonmini": json.dumps,
         "yaml": yaml.dump,
         "human": humanize
     }
@@ -300,10 +310,10 @@ class APIHelper:
     """)
 @click.option(
     "--output", "-o", default="",
-    type=click.Choice(["yaml", "json", "human", "pprint",
-                       "y", "j", "h", "p", ""]),
+    type=click.Choice(["yaml", "json", "jsonmini", "human", "pprint",
+                       "y", "j", "jm", "h", "p", ""]),
     show_choices=True,
-    help="""Override default output format.""")
+    help=f"Override default output format. {output_format_help}")
 @click.option(
     "--config-file", "-c", type=click.Path(),
     default="~/.config/synadm.yaml",
@@ -349,13 +359,9 @@ def root(ctx, verbose, batch, output, config_file):
     API's or Matrix API's. The default is 7 seconds. """)
 @click.option(
     "--output", "-o", type=click.Choice(["yaml", "json", "human", "pprint"]),
-    help="""How synadm displays data by default. 'human' gives a tabular or
-    list view depending on the fetched data. This mode needs your terminal to
-    be quite wide! 'json' displays exactly as the API responded. 'pprint' shows
-    nicely formatted json. 'yaml' is the currently recommended output format.
-    It doesn't need as much terminal width as 'human' does. Note that the
-    default output format can always be overridden by using global switch -o
-    (eg 'synadm -o pprint user list').""")
+    help=f"""How synadm displays data by default. {output_format_help} The
+    default output format can always be overridden by using the global
+    --output/-o switch (eg 'synadm -o pprint user list').""")
 @click.option(
     "--server-discovery", "-d", type=click.Choice(["well-known", "dns"]),
     help="""The method used for discovery of "the own homeserver name". Since
@@ -446,7 +452,7 @@ def config_cmd(helper, user_, token, base_url, admin_path, matrix_path,
         "format": click.prompt(
             "Default output format",
             default=output if output else helper.config.get("format", output),
-            type=click.Choice(["yaml", "json", "human", "pprint"])),
+            type=click.Choice(["yaml", "json", "jsonmini", "human", "pprint"])),
         "timeout": click.prompt(
             "Default http timeout",
             default=timeout if timeout else helper.config.get(
--- a/synadm/cli/__init__.py
+++ b/synadm/cli/__init__.py
@@ -63,6 +63,10 @@ def json_pretty(data):
     return json.dumps(data, indent=4)
 
 
+def json_minify(data):
+    return json.dumps(data, separator=(",", ":"))
+
+
 class APIHelper:
     """ API client enriched with CLI-level functions, used as a proxy to the
     client object.
@@ -71,7 +75,7 @@ class APIHelper:
     FORMATTERS = {
         "pprint": pprint.pformat,
         "json": json_pretty,
-        "minified": json.dumps,
+        "minified": json_minify,
         "yaml": yaml.dump,
         "human": humanize
     }
--- a/synadm/cli/__init__.py
+++ b/synadm/cli/__init__.py
@@ -59,14 +59,6 @@ def humanize(data):
     return str(data)
 
 
-def json_pretty(data):
-    return json.dumps(data, indent=4)
-
-
-def json_minify(data):
-    return json.dumps(data, separator=(",", ":"))
-
-
 class APIHelper:
     """ API client enriched with CLI-level functions, used as a proxy to the
     client object.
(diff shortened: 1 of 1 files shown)


JOJ0:
Finally drafting finaliszation of PR#84. Currently I have an issue when using `-o jm` , it does work with `-o jsonmini` though. Can't be much to fix:

```
(synadm38) ~/synadm (dev) $ synadm -o j version
{
    "server_version": "1.49.2",
    "python_version": "3.6.9"
}
(synadm38) ~/synadm (dev) $ synadm -o jm version
Traceback (most recent call last):
  File "/home/jojo/.venvs/synadm38/bin/synadm", line 11, in <module>
    load_entry_point('synadm', 'console_scripts', 'synadm')()
  File "/home/jojo/.venvs/synadm38/lib/python3.6/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/jojo/.venvs/synadm38/lib/python3.6/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/jojo/.venvs/synadm38/lib/python3.6/site-packages/click/core.py", line 1256, in invoke
    Command.invoke(self, ctx)
  File "/home/jojo/.venvs/synadm38/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/jojo/.venvs/synadm38/lib/python3.6/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/jojo/.venvs/synadm38/lib/python3.6/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/jojo/synadm/synadm/cli/__init__.py", line 326, in root
    helper_loaded = ctx.obj.load()
  File "/home/jojo/synadm/synadm/cli/__init__.py", line 155, in load
    self._set_formatter(self.output_format_cli)
  File "/home/jojo/synadm/synadm/cli/__init__.py", line 133, in _set_formatter
    self.log.debug("Formatter in use: %s - %s", self.output_format,
AttributeError: 'APIHelper' object has no attribute 'output_format'
(synadm38) ~/synadm (dev) $ 
(synadm38) ~/synadm (dev) $ synadm -o jsonmini version
{"server_version": "1.49.2", "python_version": "3.6.9"}
(synadm38) ~/synadm (dev) $ 
```

Will finish this later but certainly I am happy for an early review @Ascurius and @JacksonChen666 

Not sure if I like putting the "shared help text" variable on the very top in cmd/__init__ but I don't have a better idea where it should "live"

JOJ0:
I decided that the original one-line-json output mode now should be called "minified" with an abbreviation of "m". That also fixes the issue above. Ready for a final review.- Thanks.

synadm/cli/__init__.py
```diff
@@ -33,6 +33,15 @@
 from synadm import api
 
 
+output_format_help = """'human' gives a tabular or list view depending on the
+fetched data. This mode needs your terminal to be quite wide! 'json' returns
```
JacksonChen666, line 37: Sounds better to not refer to `human` output mode with "This mode"

```suggestion
fetched data. The 'human' mode needs your terminal to be quite wide! 'json' returns
```

(suggestion may need reformatting)

synadm/cli/__init__.py
```diff
@@ -446,7 +453,8 @@ def get_redacted_token_prompt(cli_token):
         "format": click.prompt(
             "Default output format",
             default=output if output else helper.config.get("format", output),
-            type=click.Choice(["yaml", "json", "human", "pprint"])),
+            type=click.Choice([
+                "yaml", "json", "minified", "human", "pprint"])),
```
JacksonChen666, line 457: (minor) Appears to be a duplicate of https://github.com/JOJ0/synadm/pull/92/files#diff-80ac2e90b8bc961d0c6426a1304211ae1d7720562d32d2b40e64f97a0d9b55ceR71-R77

synadm/cli/__init__.py
```diff
@@ -62,6 +71,7 @@ class APIHelper:
     FORMATTERS = {
         "pprint": pprint.pformat,
         "json": json_pretty,
+        "minified": json.dumps,
```
JacksonChen666, line 74: This isn't so... minified

```json
{"server_version": "1.76.0 (b=HEAD,t=v1.76.0,eafdb12,dirty)", "python_version": "3.10.10"}
```

there are spaces in between the `:` and `,` which could be removed. it might not be practically possible to make that happen because of how the formatter works (likely needs a wrapper function).

see the compact encoding section at https://docs.python.org/3/library/json.html

synadm/cli/__init__.py
JOJ0: Sorry I don't understand? Did I commit the same thing twice? I force pushed a couple of times. That the issue?

synadm/cli/__init__.py
JOJ0: Will change that. Thanks.

synadm/cli/__init__.py
JOJ0: Hm ok, well I was actually just looking for a better name that isn't too long for the "plain" json output format. I'd like to express that this is the json the api would respond "plain text, non-formatted", since  `json` is now used for the formatted json.

Hm, that puts me back to the start but well, yes, let's brainstorm again: Any suggestions how this output mode could be called? `onelinejson``, but that's loooon and ugly as well, haha ;-)

JOJ0: or just `oneline` ? The help would tell the use what it means anyway. What do you think?

(I don't want to fiddle around with really "minifying" at the moment, but if you would be up for it I won't stop you :-)

JOJ0: Oh, I should have read your suggestion before blablaing. I'm sorry! I read the "compact encoding" section. So if we would do that in a formatting_function, would you say we could call it `minified` ? If yes, I would just implement that. I like the idea!

JacksonChen666:
On February 22, 2023 12:37:38 PM UTC, J0J0 Todos ***@***.***> wrote:
>@JOJ0 commented on this pull request.
>
>
>
>> @@ -62,6 +71,7 @@ class APIHelper:
>     FORMATTERS = {
>         "pprint": pprint.pformat,
>         "json": json_pretty,
>+        "minified": json.dumps,
>
>Oh, I should have read your suggestion before blablaing. I'm sorry! I read the "compact encoding" section. So if we would do that in a formatting_function, would you say we could call it `minified` ? If yes, I would just implement that. I like the idea!
>

sure

synadm/cli/__init__.py
JOJ0: Is there anyghing to do here still, sorry I really didn't get what you mean. Please elaborate! Do you mean we could use the FORMATTERS dict to provide the choices with `click.choice`?
Docs say it needs a list: https://click.palletsprojects.com/en/8.1.x/options/#choice-options
You mean we cast the dict to a list and pass it here? Anyway, please elaborate what you ment! :-) Thanks

synadm/cli/__init__.py
JOJ0: I kind of rewrote that whole thing. Please review it again! Thanks!

synadm/cli/__init__.py
JOJ0: That is changed now. Actually I love the idea! Many thanks for that great suggestion! We now have a "real" minified mode slash scripting- slash performance-optimized mode. Neat!

synadm/cli/__init__.py
JacksonChen666: > Do you mean we could use the FORMATTERS dict to provide the choices with `click.choice`?

that's what I intended to say.

> Docs say it needs a list: https://click.palletsprojects.com/en/8.1.x/options/#choice-options
> You mean we cast the dict to a list and pass it here? Anyway, please elaborate what you ment! :-) Thanks

you can use `dict.keys()` (`dict` being the actual dictionary object) and it will give you a list containing the keys of the dictionary, which you can use to pass to `click.Choice` (if you need the keys but only the first letters, you'll at least need list comprehension)

synadm/cli/__init__.py
```diff
@@ -300,10 +314,10 @@ def generate_mxid(self, user_id):
     """)
 @click.option(
     "--output", "-o", default="",
-    type=click.Choice(["yaml", "json", "human", "pprint",
-                       "y", "j", "h", "p", ""]),
+    type=click.Choice(["yaml", "json", "minified", "human", "pprint",
+                       "y", "j", "m", "h", "p", ""]),
```
JacksonChen666, line 318: Why is an empty string an option?

synadm/cli/__init__.py
JacksonChen666: I think I'll decide against my own idea due to some reasons:
1. Every time the program is run, the values has to be computed
2. Code becomes a bit too clever, which is not really great
3. Understanding what the values will be will require jumping around the code more

JacksonChen666:
pushed some stuff to fix a couple of issues and other things.

commits of interest: 9a41943

JOJ0:
> pushed some stuff to fix a couple of issues and other things.

Please don't do that. I had done some of that already, got distracted by something else in my life and now force-pushed it to finish it. I would prefer if we try to keep a PR to one person as being the submitter and the other being the reviewer and ask whether we could change something in the source branch to prevent issues like that. If I would have like you to change or "take over" a PR I would have asked you. Thanks for your understanding.

synadm/cli/__init__.py
JOJ0: I was aksing that myself when I went over that code and recall something very very dusty in my brain that there once was a reason. It might be no reason anymore. It needs to be tested thoroughly whether or not in can be removed...

synadm/cli/__init__.py
JOJ0: I'm fine with that decision.

synadm/cli/__init__.py
```diff
@@ -33,6 +33,15 @@
...
 
 
+output_format_help = """The 'human' mode gives a tabular or list view depending
+on the fetched data, but often needs a lot of horizontal space to display
+correctly. 'json' returns formatted json. 'minified' is minified json, suitable
+for scripting purposes. 'pprint' shows a formatted output with the help of
+Python's built-in pprint module.  'yaml' is a compromise between human- and
+machine-readable output, it doesn't need as much terminal width as 'human' does
+and is the default on fresh installations."""
+
```
JOJ0, line 43: I rewrote that. Please review again! Thanks in advance!

JacksonChen666:
@JOJ0 Sorry for that, I have now put my additional changes on my fork. You can [compare my changes](https://github.com/JOJ0/synadm/compare/dev...JacksonChen666:synadm:pr-92-additional?expand=1) that I think could be useful.

Commits of interests are 9a41943aa8a1dfed0ae9de65fc9a1f4536415b29 and ba2812500f11650bcc77fd6383135f73d0a433f2.

b6a9ca634e77e8ebd9d0d5689862937c40e873e6 does not seem to have changed anything if my changes were merged onto the dev branch.

JOJ0:
> > pushed some stuff to fix a couple of issues and other things.
> 
> Please don't do that. I had done some of that already, got distracted by something else in my life and now force-pushed it to finish it. I would prefer if we try to keep a PR to one person as being the submitter and the other being the reviewer and ask whether we could change something in the source branch to prevent issues like that. If I would have like you to change or "take over" a PR I would have asked you. Thanks for your understanding.

And I guess I should add that there can be exceptions to that rule: If a contributor/PR submitter doesn't react for like a few weeks we should take over that branch and try to finish, or merge and just finish in master, but as said: Jumping in within a few hours to a source branch is just a little too early I would say. ;-)

synadm/cli/__init__.py
JacksonChen666: Ignoring as [you already requested to review it](https://github.com/JOJ0/synadm/pull/92#discussion_r1116659101) and [your force push](https://github.com/JOJ0/synadm/compare/ba2812500f11650bcc77fd6383135f73d0a433f2..a2c2f2882aee99437337429365c56652b66f49f9) didn't seem to change anything.

synadm/cli/__init__.py
```diff
@@ -33,6 +33,15 @@
 from synadm import api
 
 
+output_format_help = """The 'human' mode gives a tabular or list view depending
+on the fetched data, but often needs a lot of horizontal space to display
+correctly. 'json' returns formatted json. 'minified' is minified json, suitable
+for scripting purposes. 'pprint' shows a formatted output with the help of
+Python's built-in pprint module.  'yaml' is a compromise between human- and
```
JacksonChen666, line 40: There's an extra space

```suggestion
Python's built-in pprint module. 'yaml' is a compromise between human- and
```

synadm/cli/__init__.py
```diff
@@ -62,6 +75,7 @@ class APIHelper:
     FORMATTERS = {
         "pprint": pprint.pformat,
         "json": json_pretty,
+        "minified": json_minify,
```
JacksonChen666, line 78: Not a requested change, but you *could* use lambdas instead. Just an idea.

I've already done it on 9a41943.

synadm/cli/__init__.py
```diff
@@ -54,6 +63,10 @@ def json_pretty(data):
     return json.dumps(data, indent=4)
 
 
+def json_minify(data):
+    return json.dumps(data, separator=(",", ":"))
```
JacksonChen666, line 67: You named a parameter incorrectly, so it doesn't work.

```suggestion
    return json.dumps(data, separators=(",", ":"))
```

JacksonChen666 (changes requested):
Already reviewed this a couple of times. Seems good to me, except for that one major issue.

synadm/cli/__init__.py
JOJ0: Ok, then probably I only force pushed the change with "separator" spelled wrongly. All good, then this is done :-P

JOJ0:
> ne major issue.

Seems there was a timely overlapping things going on between us. Some of the suggested changes I had fixed and pushed already. Thanks for the review. I like the lambda suggestion. It spares some lines and still is readable enough.  I  cherry-picked your commit into here but slightly changed the commit message. Hope that's ok. Thanks again!

I think I'm going to merge this now.

PR #93, opened 2023-02-19
User list pagination & Deactivate multiple accounts based on regex

### LINKED ISSUE #56: Deactivating multiple users at once
Hello,

Out of convenience, I am missing a feature for deactivating multiple users at once, preferbly based on a given regular expression or any other kind of matching pattern.

For this to work, all users must be queried and for each it must be check if the MXIDs localpart matches the regular expression. I assume that the ```list``` command can not be used for this at its current implementation. The first problem is, that the API limits how many users are returned and as far as I know there is no way of telling the API ro return all users without limit in one request. Second, the current way of filtering for users during the ```list``` command is not exactly the same as a pattern matching, since the fitler option probably only checks wheter a user ID contains a given substring or not, which is not exactly what I am looking for.

After the query is finished, the deactivate command has to take the list of user IDs returned by the modified ```list``` command and deactivate each of the accounts.

Off course this feature is entirely optional and is not coverd directly by the API at all.

JacksonChen666:
Resolves #56

Note: This PR contains more than one change:
- User list pagination
- (Main) Deactivate multiple accounts based on regex (depends on user list pagination)

I don't suggest squashing this.

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -415,14 +415,16 @@ class SynapseAdmin(ApiRequest):
         """Yields API responses for all of the pagination.
 
         Args:
-            _limit (int): maximum number of users returned, used for pagination
-            _guests (bool): enable/disable fetching of guest users
-            _deactivated (bool): enable/disable fetching of deactivated users
-            _name (string): user name localpart to search for, see Synapse
-                admin API docs for details
-            _user_id (string): fully qualified Matrix user ID to search for
-            _from (string): offsets user list by this number, used for
-                pagination
+            _limit (int): Maximum number of users returned, used for
+                pagination.
+            _guests (bool): Enable/disable fetching of guest users.
+            _deactivated (bool): Enable/disable fetching of deactivated
+                users.
+            _name (string): User name localpart to search for, see Synapse
+                admin API docs for details.
+            _user_id (string): Fully qualified Matrix user ID to search for.
+            _from (string): Offsets user list by this number, used for
+                pagination.
 
         Yields:
             dict: The admin API response for listing accounts.
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -420,10 +420,7 @@ class SynapseAdmin(ApiRequest):
             response = self.user_list(_from, _limit, _guests, _deactivated,
                                       _name, _user_id)
             yield response
-            if "next_token" in response.keys():
-                _from = response["next_token"]
-            else:
-                _from = None
+            _from = response.get("next_token", None)
 
     def user_membership(self, user_id, return_aliases, matrix_api):
         """Get a list of rooms the given user is member of
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -411,10 +411,22 @@ class SynapseAdmin(ApiRequest):
 
     def user_list_paginate(self, _limit, _guests, _deactivated,
                            _name, _user_id, _from="0"):
+        # documentation is mostly duplicated from user_list...
         """Yields API responses for all of the pagination.
 
         Args:
-            _from (string): Opaque string
+            _limit (int): maximum number of users returned, used for pagination
+            _guests (bool): enable/disable fetching of guest users
+            _deactivated (bool): enable/disable fetching of deactivated users
+            _name (string): user name localpart to search for, see Synapse
+                admin API docs for details
+            _user_id (string): fully qualified Matrix user ID to search for
+            _from (string): offsets user list by this number, used for
+                pagination
+
+        Yields:
+            dict: The admin API response for listing accounts.
+                https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#list-accounts
         """
         while _from is not None:
             response = self.user_list(_from, _limit, _guests, _deactivated,
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -409,6 +409,22 @@ class SynapseAdmin(ApiRequest):
             "user_id": _user_id
         })
 
+    def user_list_paginate(self, _limit, _guests, _deactivated,
+                           _name, _user_id, _from="0"):
+        """Yields API responses for all of the pagination.
+
+        Args:
+            _from (string): Opaque string
+        """
+        while _from is not None:
+            response = self.user_list(_from, _limit, _guests, _deactivated,
+                                      _name, _user_id)
+            yield response
+            if "next_token" in response.keys():
+                _from = response["next_token"]
+            else:
+                _from = None
+
     def user_membership(self, user_id, return_aliases, matrix_api):
         """Get a list of rooms the given user is member of
 
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -179,7 +179,7 @@ def deactivate_regex(ctx, helper, regex, gdpr_erase, dry_run=False,
                                                             True, False, "",
                                                             ""):
         for user in list_user_response["users"]:
-            if pattern.search(user["name"]):
+            if pattern.match(user["name"]):
                 if dry_run:
                     click.echo(f"Would deactivate: {user['name']}")
                     continue
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -153,7 +153,8 @@ def deactivate(ctx, helper, user_id, gdpr_erase):
     help="""Do everything except deactivating users."""
 )
 @click.option(
-    "--paginate", type=int, default=500, show_default=True,
+    "--batch-size", "--paginate", "-p", type=int, default=500,
+    show_default=True,
     help="""How many users should be requested from the API one at a time.
     This option has no effect on how many users will be deactivated.
 
@@ -162,8 +163,7 @@ def deactivate(ctx, helper, user_id, gdpr_erase):
 )
 @click.pass_obj
 @click.pass_context
-def deactivate_regex(ctx, helper, regex, gdpr_erase, dry_run,
-                     batch_size):
+def deactivate_regex(ctx, helper, regex, gdpr_erase, dry_run, batch_size):
     """ Deactivate or GDPR-erase accounts based on regex.
 
     Does everything normal deactivation does, just for multiple users matching
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -161,8 +161,8 @@ def deactivate(ctx, helper, user_id, gdpr_erase):
 )
 @click.pass_obj
 @click.pass_context
-def deactivate_regex(ctx, helper, regex, gdpr_erase, dry_run=False,
-                     batch_size=500):
+def deactivate_regex(ctx, helper, regex, gdpr_erase, dry_run,
+                     batch_size):
     """ Deactivate or GDPR-erase accounts based on regex.
 
     Does everything normal deactivation does, just for multiple users. The
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -165,13 +165,19 @@ def deactivate_regex(ctx, helper, regex, gdpr_erase, dry_run,
                      batch_size):
     """ Deactivate or GDPR-erase accounts based on regex.
 
-    Does everything normal deactivation does, just for multiple users. The
-    --batch argument may be helpful if you do not want to be prompted to
-    deactivate every single account, and the --dry-run argument is helpful
-    to see which accounts will be deactivated.
+    Does everything normal deactivation does, just for multiple users matching
+    the given regex.
 
-    The regex argument takes a string and uses Python's regex matching. It
-    matches based on the full matrix ID."""
+    The regex argument takes a string and uses Python's re.match (matches
+    regex starting from first character). A regex is expected to match a
+    full Matrix ID, or partially at least from the first character.
+
+    --dry-run can be used to see which accounts will be deactivated. This
+    can be useful for reviewing the accounts that will be deactivated.
+
+    Additionally, the --batch argument (given before the subcommands) will
+    not prompt for if you want to deactivate a user (very useful for many
+    users)."""
     helper.log.debug(f"Regex: {regex}")
     # if below fails, turn on debug mode to get the actual given regex.
     pattern = re.compile(regex)
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -146,7 +146,8 @@ def deactivate(ctx, helper, user_id, gdpr_erase):
     help="""Marks the users as GDPR-erased. This means messages sent by the
     users will still be visible by anyone that was in the room when these
     messages were sent, but hidden from other users joining the room
-    afterwards.""", show_default=True)
+    afterwards.""", show_default=True
+)
 @click.option(
     "--dry-run", "-n", is_flag=True, default=False,
     help="""Do everything except deactivating users."""
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -151,9 +151,18 @@ def deactivate(ctx, helper, user_id, gdpr_erase):
     "--dry-run", "-n", is_flag=True, default=False,
     help="""Do everything except deactivating users."""
 )
+@click.option(
+    "--batch-size", type=int, default=500, show_default=True,
+    help="""How many users should be requested from the API one at a time.
+    This option has no effect on how many users will be deactivated.
+
+    Increasing this is not necessary in most cases but useful if you have a
+    lot of accounts on your homeserver."""
+)
 @click.pass_obj
 @click.pass_context
-def deactivate_regex(ctx, helper, regex, gdpr_erase, dry_run=False):
+def deactivate_regex(ctx, helper, regex, gdpr_erase, dry_run=False,
+                     batch_size=500):
     """ Deactivate or GDPR-erase accounts based on regex.
 
     Does everything normal deactivation does, just for multiple users. The
@@ -166,7 +175,9 @@ def deactivate_regex(ctx, helper, regex, gdpr_erase, dry_run=False):
     helper.log.debug(f"Regex: {regex}")
     # if below fails, turn on debug mode to get the actual given regex.
     pattern = re.compile(regex)
-    for list_user_response in helper.api.user_list_paginate(100, True, False, "", ""):
+    for list_user_response in helper.api.user_list_paginate(batch_size,
+                                                            True, False, "",
+                                                            ""):
         for user in list_user_response["users"]:
             if pattern.search(user["name"]):
                 if dry_run:
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -18,6 +18,7 @@
 """ User-related CLI commands
 """
 
+import re
 import click
 from click_option_group import optgroup, MutuallyExclusiveOptionGroup
 from click_option_group import RequiredAnyOptionGroup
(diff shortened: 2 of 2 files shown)


synadm/cli/user.py
```diff
@@ -138,6 +139,55 @@ def deactivate(ctx, helper, user_id, gdpr_erase):
...
+                     batch_size=500):
+    """ Deactivate or GDPR-erase accounts based on regex.
+
+    Does everything normal deactivation does, just for multiple users. The
+    --batch argument may be helpful if you do not want to be prompted to
+    deactivate every single account, and the --dry-run argument is helpful
+    to see which accounts will be deactivated.
+
+    The regex argument takes a string and uses Python's regex matching. It
+    matches based on the full matrix ID."""
```
JOJ0, line 174: Help texts are precise, not too long and to the point! That is very much appreciated! :heart: 
Because of the last sentecne "....based on the full matrix ID." see below for details but actually we are currently matching even if it's not a full matrix ID (leaving out the @ in the beginning is rather "matching part of the matrix ID). I might be too picky here with the wording so you can as well ignore that and leave it as-is, it's fine, but read on about streamlining below. In what way we need to decide but we should try to "share behaviour" between commands as good as possible.

synadm/cli/user.py
```diff
@@ -138,6 +139,55 @@ def deactivate(ctx, helper, user_id, gdpr_erase):
...
+    help="""Marks the users as GDPR-erased. This means messages sent by the
+    users will still be visible by anyone that was in the room when these
+    messages were sent, but hidden from other users joining the room
+    afterwards.""", show_default=True)
+@click.option(
+    "--dry-run", "-n", is_flag=True, default=False,
+    help="""Do everything except deactivating users."""
+)
+@click.option(
+    "--batch-size", type=int, default=500, show_default=True,
```
JOJ0, line 155: Does it make sense to set this value to same as the notice send command does? (Considering once we rewrite notice send to use your new pagination solution)

synadm/cli/user.py
```diff
@@ -138,6 +139,55 @@ def deactivate(ctx, helper, user_id, gdpr_erase):
...
+    The regex argument takes a string and uses Python's regex matching. It
+    matches based on the full matrix ID."""
+    helper.log.debug(f"Regex: {regex}")
+    # if below fails, turn on debug mode to get the actual given regex.
+    pattern = re.compile(regex)
+    for list_user_response in helper.api.user_list_paginate(batch_size,
+                                                            True, False, "",
+                                                            ""):
+        for user in list_user_response["users"]:
+            if pattern.search(user["name"]):
```
JOJ0, line 182: I think we have a difference here to the notice send command. See the examples below to see what I mean. 
We've used re.match instead of re.search.

Actually I like the behaviour with your solution in this PR better, but I recall we had some concerns back when we reviewed the notice send feature and thus changed it so it explicitely matches from the beginning (requires the @ or .* to match). At the end of the day we should streamline both commands as good and sensible as possible. Probably better we move in the direction of a behaviour as you suggest here and later on refactor notice send to use what we learnt here. 
If something doesnt work out with the notice send command. we can still change deactivate-regex again. Well, only if we do the refactor soonish, before next release.


```
(synadm38) ~/git/jacksonchen666_synadm (deactivate-regex) $ synadm user deactivate-regex testuser -n
Would deactivate: @testuser1:peek-a-boo.at
Would deactivate: @testuser2:peek-a-boo.at
Would deactivate: @testuser3:peek-a-boo.at
Would deactivate: @testuser4:peek-a-boo.at
Would deactivate: @testuser5:peek-a-boo.at
Would deactivate: @testuser6:peek-a-boo.at
Would deactivate: @testuser:peek-a-boo.at
(synadm38) ~/git/jacksonchen666_synadm (deactivate-regex) $ 
(synadm38) ~/git/jacksonchen666_synadm (deactivate-regex) $ synadm notice send -r @testuser msg 

Note: When sending to a large amount of recipients, consider using the --silent option.

Recipients:
 - @testuser1:peek-a-boo.at
 - @testuser2:peek-a-boo.at
 - @testuser3:peek-a-boo.at
 - @testuser4:peek-a-boo.at
 - @testuser5:peek-a-boo.at
 - @testuser6:peek-a-boo.at
 - @testuser:peek-a-boo.at

Plaintext message:
---
msg
---
Formatted message:
---
msg
---
Send now? [y/N]: 
(synadm38) ~/git/jacksonchen666_synadm (deactivate-regex) $ 
(synadm38) ~/git/jacksonchen666_synadm (deactivate-regex) $ synadm notice send -r testuser msg 

Note: When sending to a large amount of recipients, consider using the --silent option.

Recipients:
(no recipient matched)

Plaintext message:
---
msg
---
Formatted message:
---
msg
---
Send now? [y/N]: 
(synadm38) ~/git/jacksonchen666_synadm (deactivate-regex) $ synadm notice send -r '.*testuser' msg 

Note: When sending to a large amount of recipients, consider using the --silent option.

Recipients:
 - @testuser1:peek-a-boo.at
 - @testuser2:peek-a-boo.at
 - @testuser3:peek-a-boo.at
 - @testuser4:peek-a-boo.at
 - @testuser5:peek-a-boo.at
 - @testuser6:peek-a-boo.at
 - @testuser:peek-a-boo.at

Plaintext message:
---
msg
---
Formatted message:
---
msg
---
Send now? [y/N]: 
(synadm38) ~/git/jacksonchen666_synadm (deactivate-regex) $
```

synadm/cli/user.py
```diff
@@ -138,6 +139,55 @@ def deactivate(ctx, helper, user_id, gdpr_erase):
...
+    "--batch-size", type=int, default=500, show_default=True,
+    help="""How many users should be requested from the API one at a time.
+    This option has no effect on how many users will be deactivated.
+
+    Increasing this is not necessary in most cases but useful if you have a
+    lot of accounts on your homeserver."""
+)
+@click.pass_obj
+@click.pass_context
+def deactivate_regex(ctx, helper, regex, gdpr_erase, dry_run=False,
```
JOJ0, line 164: I was brainstorming whether this command shouldn't be called something that says "deactivate multiple accounts" but in the end I didn't come up with any good idea since `deactivate-multiple` is just longer and I don't like it. So yes, probably `deactivate-regex` is the best option and it actually tells from the beginning that the user should be aware of how to use regex as well as understand that it _could_ mean multiple. :-)

JOJ0 (commented):
Great addition to synadm and I'm very happy we finally have a neat implementation for pagination ourselves. It will come in handy in the future often!! Thanks! Great work!

synadm/api.py
```diff
@@ -409,6 +409,22 @@ def user_list(self, _from, _limit, _guests, _deactivated,
...
+        """
+        while _from is not None:
+            response = self.user_list(_from, _limit, _guests, _deactivated,
+                                      _name, _user_id)
+            yield response
+            if "next_token" in response.keys():
+                _from = response["next_token"]
+            else:
+                _from = None
+
```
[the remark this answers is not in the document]
JacksonChen666: > We try to follow the google docstrings specifications/suggestions throughout the project so far.

What document should be used as a reference for the "google docstrings"? I found [this](https://google.github.io/styleguide/pyguide.html), and I think a CONTRIBUTING.md file explaining how functions should be documented could be helpful (along with many other things you'd like to have including some things I missed in my first PR).

> Could you look up what they'd recommend for generator functions like this? Thanks!

`Yields` instead of `Returns`. https://google.github.io/styleguide/pyguide.html#s2.9-generators

synadm/cli/user.py
JacksonChen666: I chose 500 cause that's a large number. It was 100 at one point, but I made it higher since you know, there could be a lot of users even on a single user homeserver (I have 7186 users on my HS).

So, in the case of 500, 15 user list requests are made. In the case of 100, 72 user list requests are made.
500 isn't a *lot*, but 100 is *even less*.

It makes more sense to have a higher batch size/pagination size in cases where computation on each user isn't a lot.

(formula for calculating amount of requests needed to paginate all users: `ceil(users/batch_size)`)

synadm/cli/user.py
JacksonChen666: I think having the regex `match` requires the admin to know what they're doing a bit more, whereas using `search` means you can give "spambot" as a regex and have it match based on whether the matrix ID contains "spambot".

`match` could be inconvenient, but `search` might be easy to screw up. Both are regex, so full proper regexes work either way.

Of course, now it's the question of "which one cause we don't want to be inconsistent". I would lean towards `match`.

Search based:
```
> synadm user deactivate-regex -n 42
Would deactivate: @142:localhost:8480
Would deactivate: @242:localhost:8480
Would deactivate: @342:localhost:8480
Would deactivate: @420:localhost:8480
Would deactivate: @421:localhost:8480
Would deactivate: @422:localhost:8480
Would deactivate: @423:localhost:8480
Would deactivate: @424:localhost:8480
Would deactivate: @425:localhost:8480
Would deactivate: @426:localhost:8480
Would deactivate: @427:localhost:8480
Would deactivate: @428:localhost:8480
Would deactivate: @429:localhost:8480
Would deactivate: @42:localhost:8480
Would deactivate: @442:localhost:8480
Would deactivate: @542:localhost:8480
Would deactivate: @642:localhost:8480
> synadm user deactivate-regex -n '@.*42.*?:.*'
Would deactivate: @142:localhost:8480
Would deactivate: @242:localhost:8480
Would deactivate: @342:localhost:8480
Would deactivate: @420:localhost:8480
Would deactivate: @421:localhost:8480
Would deactivate: @422:localhost:8480
Would deactivate: @423:localhost:8480
Would deactivate: @424:localhost:8480
Would deactivate: @425:localhost:8480
Would deactivate: @426:localhost:8480
Would deactivate: @427:localhost:8480
Would deactivate: @428:localhost:8480
Would deactivate: @429:localhost:8480
Would deactivate: @42:localhost:8480
Would deactivate: @442:localhost:8480
Would deactivate: @542:localhost:8480
Would deactivate: @642:localhost:8480
> synadm user deactivate-regex -n @42
Would deactivate: @420:localhost:8480
Would deactivate: @421:localhost:8480
Would deactivate: @422:localhost:8480
Would deactivate: @423:localhost:8480
Would deactivate: @424:localhost:8480
Would deactivate: @425:localhost:8480
Would deactivate: @426:localhost:8480
Would deactivate: @427:localhost:8480
Would deactivate: @428:localhost:8480
Would deactivate: @429:localhost:8480
Would deactivate: @42:localhost:8480
```

Match based:
```
> synadm user deactivate-regex -n 42
> synadm user deactivate-regex -n '@.*42.*?:.*'
Would deactivate: @142:localhost:8480
Would deactivate: @242:localhost:8480
Would deactivate: @342:localhost:8480
Would deactivate: @420:localhost:8480
Would deactivate: @421:localhost:8480
Would deactivate: @422:localhost:8480
Would deactivate: @423:localhost:8480
Would deactivate: @424:localhost:8480
Would deactivate: @425:localhost:8480
Would deactivate: @426:localhost:8480
Would deactivate: @427:localhost:8480
Would deactivate: @428:localhost:8480
Would deactivate: @429:localhost:8480
Would deactivate: @42:localhost:8480
Would deactivate: @442:localhost:8480
Would deactivate: @542:localhost:8480
Would deactivate: @642:localhost:8480
> synadm user deactivate-regex -n @42
Would deactivate: @420:localhost:8480
Would deactivate: @421:localhost:8480
Would deactivate: @422:localhost:8480
Would deactivate: @423:localhost:8480
Would deactivate: @424:localhost:8480
Would deactivate: @425:localhost:8480
Would deactivate: @426:localhost:8480
Would deactivate: @427:localhost:8480
Would deactivate: @428:localhost:8480
Would deactivate: @429:localhost:8480
Would deactivate: @42:localhost:8480
```

synadm/api.py
```diff
@@ -409,6 +409,22 @@ def user_list(self, _from, _limit, _guests, _deactivated,
...
+        Args:
+            _from (string): Opaque string
+        """
+        while _from is not None:
+            response = self.user_list(_from, _limit, _guests, _deactivated,
+                                      _name, _user_id)
+            yield response
+            if "next_token" in response.keys():
+                _from = response["next_token"]
```
JacksonChen666, line 424: because of how clever this code is, the `_from` variable is modified. this can have undesirable effects for callers passing their valuable variables:
```
>>> def asdf(_asdf):
...     _asdf += ["i am not supposed to be in your var"]
...     return _asdf
...
>>> a = ['hi']
>>> a
['hi']
>>> asdf(a)
['hi', 'i am not supposed to be in your var']
>>> a
['hi', 'i am not supposed to be in your var']
```

this applies to lists (which are not expected) and maybe other things too.

synadm/cli/user.py
JacksonChen666: Additionally, streamlining the output of `deactivate-regex -n` to be similar to sending multiple notices would be kind of... odd, since `deactivate-regex -n` outputs the list of users to deactivate live.

You can see it happening: https://asciinema.org/a/Vj461YV05mmhAof7bUe53y9BQ

(Also, are we putting these messages on stdout? Should be stderr instead cause we definitely do not want to mess with stdout containing human messages requesting interaction when the user expected the stdout to be json or the format they chose)

synadm/api.py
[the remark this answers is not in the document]
JOJ0: Sorry I was too lazy to google for google myself ;-) Yes, you found the right thing! And yes, a contributing.md is a missing piece in the project and I will try to work on that soonish.

synadm/cli/user.py
JOJ0: Thanks for elaborating! So regarding "computation on each user isn't a lot" we could say that `notice send` and `user deactivate` is actually kind of the same. We send one post query to Synapse for each user. So in your case 500 times in a row `deactivate` and in the other command 100 times `send notice`. I don't really know what the computational difference for a Synapse server would be. Maybe we set the `notice` command to 500 as well? What do you think?

And thanks for that formula, I now get how much computation is required for "pagination", so actually it's better to not set too tiny pages.

Hope I got all that right! :sweat_smile: 

So because of streamlining usability: What remains is that here we use `--batch-size` and over there at `notice? command we use `--paginate`. I really would like to find use the same name for that. I don't have a hard preference which one it should be as long as we streamline it. What do you think?

synadm/cli/user.py
JOJ0: Let's streamline the output of both commands "as good and as sensible as possible". The `notice send` command preview is very special and certainly is a different approach to what `deactivate-regex -n` does. It didn't mean to streamline in that regard! I like how you designed `-n`. Keep it that way please! :-)

JOJ0: > (Also, are we putting these messages on stdout? Should be stderr instead cause we definitely do not want to mess with stdout containing human messages requesting interaction when the user expected the stdout to be json or the format they chose)

That is certainly a valid point but up to now we don't have any distinguishing between what output of data and "informational messages" is. What we have though is the `--yes/--batch` global flag. It should always only output data, **as long as it's not errors**.

Certainly sending particular things to stderr instead of stdout would be the right thing to do but that's a whole new chapter. I would like to continue that discussion but I suppose it will be quite some work to refactor all parts of synadm to follow that pattern. But again: Yes it would be the right thing to do for a perfect unix cli tool! :-)

synadm/api.py
```diff
@@ -411,10 +411,22 @@ def user_list(self, _from, _limit, _guests, _deactivated,
...
         Args:
-            _from (string): Opaque string
+            _limit (int): maximum number of users returned, used for pagination
+            _guests (bool): enable/disable fetching of guest users
+            _deactivated (bool): enable/disable fetching of deactivated users
+            _name (string): user name localpart to search for, see Synapse
+                admin API docs for details
+            _user_id (string): fully qualified Matrix user ID to search for
+            _from (string): offsets user list by this number, used for
+                pagination
```
JOJ0, line 425: Sorry for being picky: I feel that starting statements after the `:` with capital letters improves human-readability (even though it's not technically correct since sometimes they are not full sentences). Thanks!

synadm/api.py
```diff
@@ -411,10 +411,22 @@ def user_list(self, _from, _limit, _guests, _deactivated,
...
+
+        Yields:
+            dict: The admin API response for listing accounts.
+                https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#list-accounts
```
JOJ0, line 429: I kind of like the idea of documenting by linking to the actual API docs. I'd like to continue discussing that idea for  throughout synadm in#synadm. I did something like that with a general statement on the very top of  api.py IIRC but let's discuss if we would like to explicitely state a link in every API method.

synadm/cli/user.py
JOJ0: Also I see you decided for regex `match`. Ok with me even though I started to like the idea of having the more flexible `search` (I always forget to put @ in the beginning of my regex myself and wonder for a second ;-)). Anyway I think for now `match` is the easier to solution since if we would use `search` here we would need to test `notice send` again, and research why this decision was made. I'm not sure anymore but we talked about it a lot back then. So, yes, let's leave `match`, streamlining is always to be preferred IMHO!

JacksonChen666:
...not sure how this PR stalled

JOJ0:
> ...not sure how this PR stalled

Some change requests here: https://github.com/JOJ0/synadm/pull/93#discussion_r1116711271

Also unanswered questions around how the streamlining with --paginate (notice send cmd) and --batch-size should be solved.

Thanks!

JacksonChen666:
> Also unanswered questions around how the streamlining with --paginate (notice send cmd) and --batch-size should be solved.

I think `--batch-size` makes sense, but it's also an option that doesn't have to be touched by the user most of the time.

About `--paginate`, it's not super obvious what it means. `--batch-size` is a little more obvious which is why I chose it.

JOJ0:
> > Also unanswered questions around how the streamlining with --paginate (notice send cmd) and --batch-size should be solved.
> 
> `--paginate` is not super obvious about what is does. `--batch-size` is a little more obvious (to me) which is why I chose it.
> 
> (Are we discussion option names and assuming the user won't use the help function? Maybe)

We are discussing that it is bad UI design if the same thing is named differently throughout the project.

Certainly we assume that our valued users will use `--help`.

We at least try to achieve a "cohesive experience" when it is possible.

JacksonChen666:
I'm not sure what the option is suggested to be, so I changed `batch-size` to `paginate` to be consistent with the notice send command.

JOJ0:
> I'm not sure what the option is suggested to be, so I changed `batch-size` to `paginate` to be consistent with the notice send command. Feel free to `git revert` that commit if you don't want it like that but instead as `batch-size`.

Honestly, I would prefer option `--batch-size` because as you say it definitely is very self-explanatory and thus a perfect choice IMO but I would leave the final choice to you since if we go down that route of strealining, there is some things to consider:

- The original submitter of the `notice send` command, IIRC correctly, decided for `--paginate`. IIRC there was some discussion around easily being confusable with the global option `--batch` if it would be used for something like batch size of messages to send out at once. It was in the very early "design phase" where not even a PR was existing yet and it was brainstormed how a command for sending "server notices" could look like. It was a lively and productive discussion  but probably hard to follow for reading through it. Just in case you're interested: https://matrix.to/#/!mLATeUxylgHiofUzHJ:peek-a-boo.at/$wEevQ52VnvV7idh8H-WMsIhu5g4Ge3gLIBsVjaUA0kY?via=jacksonchen666.com&via=matrix.org&via=envs.net
- Anyway, the submitter decided for `--paginate`, while in the meantime _I_ decided that alternative names for `--batch` would be good which I brought in with #77 
- As I see it now the confusion of a global `--batch/--non-interactive` option with a subcommand-dedicated `--batch-size` option is not too big. Only thing to be careful is probably to not name something just `batch` in the subcommand's code since we use that for the global flag often.

So long story short I would appreciate if you would go down the road of naming it `--batch-size` but not forget to change this in subsequent PR for `notice send` as well.

I would like to start a discussion in #synadm:peek-a-boo.at whether or not people think that such a change would be offending to them.

An alternative option would be to just pragmatically add both options `--batch-size` as well as `--paginate`. In a subsequent PR or actually probably it would work if just being added to this one: Add `--batch-size` as an alias to the `notice send` command as well.#

Sorry for the length but sometimes making decisions is not as easy as it seems. There you have some backstory and I hope it helps moving on in the right direction.

JOJ0:
Hi @JacksonChen666, is it ok for you if I hijack this PR/branch? I would like to have both options: `--batch-size` and `--paginate`.

I have a pending change that tries to reduce a possible confusing with what we called "batch mode" so far in PR #102

JacksonChen666:
sure, you can hijack the PR for the options.

synadm/cli/user.py
```diff
@@ -138,6 +139,62 @@ def deactivate(ctx, helper, user_id, gdpr_erase):
...
+    show_default=True,
+    help="""How many users should be requested from the API one at a time.
+    This option has no effect on how many users will be deactivated.
+
+    Increasing this is not necessary in most cases but useful if you have a
+    lot of accounts on your homeserver."""
+)
+@click.pass_obj
+@click.pass_context
+def deactivate_regex(ctx, helper, regex, gdpr_erase, dry_run, batch_size):
```
JOJ0, line 166: After re-adding batch_size, the forgotten renaming of this option is not required anymore since I readded `--batch-size` in the first position which makes it the variable name `batch_size`

JOJ0 (approved):
Alright, I'm just reviewing this quickly myself. Die some playing around / testing with the commands. Hope I didn't miss anything but I think we are good to go.

synadm/cli/user.py
JacksonChen666: ?

JOJ0: You changed the option name in your last commit to paginate. That wouldn't have been enough. It wouldn't have worked.

JOJ0: No problem, just saying that it wouldn't have worked but now all good :-) No worries!!

JacksonChen666: > You changed the option name in your last commit to `--paginate` but forgot to change the last argument to `paginate`.

ah, understood.

PR #98, opened 2023-03-02
Adapt Click version requirement

### LINKED ISSUE #97: synadm 0.38 requires Click<8.0,>=7.0, but you have click 8.1.3 which is incompatible.
Hello,
I'm getting this while trying to:

`pip3 install synadm`

I need Flask installed too and that requires `click>=8.0`

JOJ0:
from setup.py and requirements.txt and add missing wheel package to fix pyyaml bdist_wheel error during pip install.

Fixes #97

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,8 +54,7 @@ setup(
         "tabulate",
         "PyYaml",
         "click-option-group>=0.5.2",
-        "dnspython",
-        "wheel"
+        "dnspython"
     ],
     entry_points="""
         [console_scripts]
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ setup(
     ],
     packages=find_packages(),
     install_requires=[
-        "Click",
+        "Click>=7.0,<9.0",
         "requests",
         "tabulate",
         "PyYaml",
--- a/setup.py
+++ b/setup.py
@@ -49,12 +49,13 @@ setup(
     ],
     packages=find_packages(),
     install_requires=[
-        "Click>=7.0,<8.0",
+        "Click",
         "requests",
         "tabulate",
         "PyYaml",
         "click-option-group>=0.5.2",
-        "dnspython"
+        "dnspython",
+        "wheel"
     ],
     entry_points="""
         [console_scripts]


JacksonChen666 (commented):
warning: many nitpicks.

---

https://github.com/JOJ0/synadm/pull/98/commits/ad4aa1c89f61f1a8d2b0478af4b8cddf294d51ba does not summarize all changes in the commit summary.
"Remove Click version requirement" does not include the addition of the wheel package.

do we really need the wheel package ourselves? why do other project include/not include the wheel package? should the problematic package require wheel instead?

---

concern: the major versions is used to indicate breaking changes in compliant semver versioning (did not yet verify if Click is semver compliant).

in Click 8.0.0, python version 2 and 3.5 support was dropped. that's a breaking change, so that increments the major version.

if Click makes a breaking change (which hopefully increments the major version) and we don't notice, things could break in production and we might not know about it until it's too late (broken things reaches to users).

JOJ0:
> warning: many nitpicks.

as we are used from you lol. It's fine and appreciated :-)



> https://github.com/JOJ0/synadm/pull/98/commits/ad4aa1c89f61f1a8d2b0478af4b8cddf294d51ba does not summarize all changes in the commit summary.
> 
> "Remove Click version requirement" does not include the addition of the wheel package.

U did read my whole commit message, didnt u? Why is it important to have that in the headline of the commit as well?


> do we really need the wheel package ourselves? why do other project include/not include the wheel package? should the problematic package require wheel instead?

That is exactly the question I was asking myself and why I openend this PR.


> 
> concern: the major versions is used to indicate breaking changes in compliant semver versioning (did not yet verify if Click is semver compliant).
> 
> 
> 
> in Click 8.0.0, python version 2 and 3.5 support was dropped. that's a breaking change, so that increments the major version.

2 and 3.5 shouldnt concern us

> 
> 
> if Click makes a breaking change (which hopefully increments the major version) and we don't notice, things could break in production and we might not know about it until it's too late (broken things reaches to users).

so you suggest we should better give at least any version tequirement? I suggest Click<9 and let's research if they are semver compatible!

Thanks so far!

JacksonChen666:
> > [ad4aa1c](https://github.com/JOJ0/synadm/commit/ad4aa1c89f61f1a8d2b0478af4b8cddf294d51ba) does not summarize all changes in the commit summary.
> > "Remove Click version requirement" does not include the addition of the wheel package.
> 
> U did read my whole commit message, didnt u? Why is it important to have that in the headline of the commit as well?

the commit summary (first line) should be short (50 or 72 characters or something) and describe the majority of the changes in the commit. if the commit summary is too long, try committing a smaller change.
following that guideline makes for more smaller commits which can be helpful with narrowing down the commit that caused an issue with [git bisect](https://git-scm.com/docs/git-bisect), as well as avoiding sneaky changes.

> > concern: the major versions is used to indicate breaking changes in compliant semver versioning (did not yet verify if Click is semver compliant).
> > in Click 8.0.0, python version 2 and 3.5 support was dropped. that's a breaking change, so that increments the major version.
> 
> 2 and 3.5 shouldnt concern us

well, it sure doesn't concern this project. that should still be a demonstration of a breaking change because of click incrementing the major version (because of a breaking change).

> > if Click makes a breaking change (which hopefully increments the major version) and we don't notice, things could break in production and we might not know about it until it's too late (broken things reaches to users).
> 
> so you suggest we should better give at least any version tequirement? I suggest Click<9 and let's research if they are semver compatible!

we should probably have *some* limit on what versions are allowed (minimum usable version and maximum major version), so we don't have compatibility issues when the user somehow manages to install click v1.0.0 and that breaks everything.

JOJ0:
> > > in Click 8.0.0, python version 2 and 3.5 support was dropped. that's a breaking change, so that increments the major version.
> 
> > 
> 
> > 2 and 3.5 shouldnt concern us
> 
> 
> 
> well, it sure doesn't concern this project. that should still be a demonstration of a breaking change because of click incrementing the major version (because of a breaking change).

I now understand what you ment. Agreed. We should stick to a max major, currently I'd say between 7.x and 8.x

> 
> 
> > 
> 
> > so you suggest we should better give at least any version tequirement? I suggest Click<9 and let's research if they are semver compatible!
> 
> 
> 
> we should probably have *some* limit on what versions are allowed (minimum usable version and maximum major version), so we don't have compatibility issues when the user somehow manages to install click v1.0.0 and that breaks everything.

Agreed, will push and release that tomorrow.

Also I decided: wheel package shouldnt concern us.The error is actually a warning. You are right probably that's more of upstream's business. We could open an issue with them. Looked quickly and might have found something open already...

JacksonChen666 (approved):

PR #102, opened 2023-03-08
Rename from batch mode to non-interactive mode

JOJ0:
- Describe the "mode" in --help as "non-interactive mode"
- Add a new alias for the mode named "--no-confirm"
- Use this name as the variable name throught the project's code.
- Remove the alias "--scripting" of that mode. This might be a breaking change for users that should be discussed before merging.

### DIFF
--- a/synadm/cli/__init__.py
+++ b/synadm/cli/__init__.py
@@ -83,10 +83,10 @@ class APIHelper:
         "homeserver": "auto-retrieval"
     }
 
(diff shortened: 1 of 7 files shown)


JacksonChen666 (approved):
looks good to me:
- no more occurrences of "batch" (at least related to the option)
- `room delete` with the `--non-interactive` flag works

JOJ0:
Decided to keep `--scripting` alias. Merging.

PR #103, opened 2023-03-08
Fix changing admin-enabled user to regular user

### LINKED ISSUE #100: User modify API is poorly defined and only half works
> https://github.com/gergelypolonkai/synadm/blob/dba8132dd74fe4be1129a78e07dd26806a6331d8/synadm/api.py#L536-L561
> 
> `None` is not considered truthy, so [using `None` as "default"](https://github.com/JOJ0/synadm/pull/90/files#diff-a41082dbe9c56e01f08c3854a53edc1f26bae33f9de966cd4ed8e228fc2bf621R425) would cause the values to *never* change. This makes for a poorly functional API that is also very poorly defined (the severe lack of docs is included).
>
> Basically, we have a problem that has nothing to do with this PR, but the PR has the same problems.
> We can merge this PR without fixing the issue, but we definitely need to fix the issue which is more than just user types (trying to remove admin doesn't work because of a similar issue (`False` will not pass an implicit if condition)).

The user modify API and command works together poorly, unable to remove admin rights (too lazy to write that word). This is because it's using implicit if conditions, where values of `False` and `None` is used as defined values, but is interpreted as non-defined values.

Example: you can give admin, not remove. you can set user type (#90), you can't revert to default user type.

_Originally posted by @JacksonChen666 in https://github.com/JOJ0/synadm/issues/90#issuecomment-1457878104_

JOJ0:
Details see commit msg. (Quick)fixes #100

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -550,7 +550,7 @@ class SynapseAdmin(ApiRequest):
             ]})
         if avatar_url:
             data.update({"avatar_url": avatar_url})
-        if admin:
+        if admin is not None:
             data.update({"admin": admin})
         if deactivation == "deactivate":
             data.update({"deactivated": True})


JOJ0:
Since this is a rather nasty bug, I'll merge this now without collecting a review. Hope that is understandable.

JacksonChen666:
I can't use the review feature anymore (cause it's merged) but it works so it's OK.

JacksonChen666:
Removed requests because I can't exactly send a review.

PR #104, opened 2023-03-09
Streamline notice send --batch-size option

JOJ0:
- Add it as the first option so the variables in code is now named batch_size and adapt code accordingly.
- Keep paginate as an alternative option.
- Change helptext just slightly.


JacksonChen666 (approved):
looks good to me (tangent sidenote: i think the batch size might not be applying because of line 138 or something else, only found 8 user list requests when it should've been 35 when batch size was 20 with 699 accounts)

JOJ0:
Let's merge this as is and remember that this feature definitely needs verification and potential fixing! Nice catch!

PR #105, opened 2023-03-12
URL encoding in query function and refactor

### LINKED ISSUE #96: Special internal room IDs breaks some commands
A user [could not delete a room](https://matrix.to/#/%23synadm%3Apeek-a-boo.at/%24LcH3oCnbiOaVRcMvPZp11fT0SruL1rjfpWd8MVVpXLw?via=jacksonchen666.com&via=maclemon.at&via=matrix.org&via=raim.ist) because the internal room ID was very special:

`!test/v7 🐈️:maunium.net` (alias thingy is `#test/v7:maunium.net`)

The `/` and the emoji should be URL encoded (or just everything special) so that synapse doesn't misinterpret the `/` and the emoji doesn't cause weird issues.
That way, even if the room ID contains special characters, it should "just work" without having to resort to a workaround (manually URL encoding the room ID, so it's `%21test%2Fv7%20%F0%9F%90%88%EF%B8%8F%3Amaunium.net` instead).

[Solution for user ref](https://matrix.to/#/%23synadm%3Apeek-a-boo.at/%24LPTAauZIkJMAEHppw3GRQq0A-iGHc5WE92NimwT5OP4?via=jacksonchen666.com&via=maclemon.at&via=matrix.org&via=raim.ist), [Handy URL encode code snippet](https://stackoverflow.com/questions/40557606/how-to-url-encode-in-python-3/54135995#54135995)
Given the code needed to URL encode in python, I would put this label an "easy issue" but I'm not sure what changes are needed to make sure room IDs are URL encoded (there probably is also multiple places to also handle it sooooo).
Related issues in another project: https://github.com/Awesome-Technologies/synapse-admin/issues/111

JacksonChen666:
1. query now accepts other things as part of str.format
2. refactor all use of the query function in the project
3. function documentation could be much better
4. needs documentation on how to use query function

resolves #96

---

Things to look out for in this PR:
- Nothing should an f-string or already have the information embedded in the URL part (AKA it should be [this](https://github.com/JOJ0/synadm/pull/105/files#diff-21da8274044fdb3c233e057a58a867c3e934030edd102cbe0a79a61858386d20R584) not [this](https://github.com/JOJ0/synadm/pull/105/files#diff-21da8274044fdb3c233e057a58a867c3e934030edd102cbe0a79a61858386d20L564))
- Basically given enough information to get back a string OR `str.format` doesn't break because it wasn't given enough things

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -74,15 +74,17 @@ class ApiRequest:
         self.verify = verify
 
     def query(self, method, urlpart, params=None, data=None, token=None,
-              base_url_override=None, verify=None):
+              base_url_override=None, verify=None, *args, **kwargs):
         """Generic wrapper around requests methods.
 
-        Handles requests methods, logging and exceptions.
+        Handles requests methods, logging and exceptions, and URL encoding.
 
         Args:
             urlpart (string): The path to the API endpoint, excluding
                 self.base_url and self.path (the part after
-                proto://fqdn:port/path).
+                proto://fqdn:port/path). It will be passed to Python's
+                str.format, so the string should not be already formatted
+                (as f-strings or with str.format) as to sanitize the URL.
             params (dict, optional): URL parameters (?param1&paarm2).  Defaults
                 to None.
             data (dict, optional): Request body used in POST, PUT, DELETE
@@ -91,6 +93,10 @@ class ApiRequest:
                 on initialization can be overwritten using this argument.
             verify(bool): Mandatory SSL verification is turned on by default
                 and can be turned off using this method.
+            *args: Arguments that will be URL encoded and passed to Python's
+                str.format.
+            **kwargs: Keyword arguments that will be URL encoded (only the
+                values) and passed to Python's str.format.
 
         Returns:
             string or None: Usually a JSON string containing
@@ -99,6 +105,14 @@ class ApiRequest:
                 JSON strings. On exceptions the error type and description are
                 logged and None is returned.
         """
+        args = list(args)
+        kwargs = dict(kwargs)
+        for i in range(len(args)):
+            args[i] = urllib.parse.quote(args[i], safe="")
+        for i in kwargs.keys():
+            kwargs[i] = urllib.parse.quote(kwargs[i], safe="")
+        urlpart = urlpart.format(*args, **kwargs)
+
         if base_url_override:
             self.log.debug("base_url override!")
             url = f"{base_url_override}/{self.path}/{urlpart}"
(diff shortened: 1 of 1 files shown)


synadm/api.py
```diff
@@ -70,15 +70,17 @@ def __init__(self, log, user, token, base_url, path, timeout, debug):
...
-        Handles requests methods, logging and exceptions.
+        Handles requests methods, logging and exceptions, and URL encoding.
 
         Args:
             urlpart (string): The path to the API endpoint, excluding
                 self.base_url and self.path (the part after
-                proto://fqdn:port/path).
+                proto://fqdn:port/path). It will be passed to Python's
+                str.format, so the string should not be already formatted
+                (as f-strings or with str.format) as to sanitize the URL.
```
JOJ0, line 83: I think your documentation is good like that now!

synadm/api.py
```diff
@@ -87,6 +89,10 @@ def query(self, method, urlpart, params=None, data=None, token=None,
                 on initialization can be overwritten using this argument.
             verify(bool): Mandatory SSL verification is turned on by default
                 and can be turned off using this method.
+            *args: Arguments that will be URL encoded and passed to Python's
+                str.format.
+            **kwargs: Keyword arguments that will be URL encoded (only the
+                values) and passed to Python's str.format.
```
JOJ0, line 95: :+1:

synadm/api.py
```diff
@@ -297,7 +311,8 @@ def room_get_id(self, room_alias):
                 response as it might contain Synapse's error message.
         """
         room_directory = self.query(
-            "get", f"client/r0/directory/room/{urllib.parse.quote(room_alias)}"
+            "get", "client/r0/directory/room/{room_alias}",
+            room_alias=room_alias
```
JOJ0, line 315: I think it would be good if we state this new concept somewhere in the contributing docs. Telling the contributors about the new handling of args. I'm not quite sure where this should be.

Around here a new chapter around string formatting maybe? https://github.com/JOJ0/synadm/blob/master/CONTRIBUTING.md#command-design

JOJ0 (changes requested):
Thanks a lot for this! LGTM except some additional documentation ideas within.

And one more thing about contributing docs but **that let be my concern**, you shouldn't be bothered with it! :-)

We need to update the [Implementation Examples chapter](https://github.com/JOJ0/synadm/blob/master/CONTRIBUTING.md#implementation-examples), it is based on very old commits and just not the reality anymore. Now that f-strings are not allowed to use anymore it is even more wrong. I will take care of fixing that!

So as mentioned within the code, your part would be to add a tiny new chapter about the new string formatting concept. Thanks a lot!

synadm/api.py
JacksonChen666: not ready to merge (conflicts and outdated dev branch mess), but i'll hand off documentation to you then.

JacksonChen666:
oops included my extra stuff, re-force pushing...

JacksonChen666:
should be good to merge

JOJ0 (approved):

Issue #110, opened 2023-06-13
regtok delete is broken


andreiz:
I'm using 0.41.2 (git clone of the repo). Running `synadm regtok delete <token>` results in:

```
Traceback (most recent call last):
  File "/root/code/synadm/.venv/bin/synadm", line 33, in <module>
    sys.exit(load_entry_point('synadm', 'console_scripts', 'synadm')())
  File "/root/code/synadm/.venv/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/root/code/synadm/.venv/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/root/code/synadm/.venv/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/root/code/synadm/.venv/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/root/code/synadm/.venv/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/root/code/synadm/.venv/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/root/code/synadm/.venv/lib/python3.10/site-packages/click/decorators.py", line 38, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/root/code/synadm/synadm/cli/regtok.py", line 146, in regtok_delete
    response = helper.api.regtok_delete(token)
  File "/root/code/synadm/synadm/api.py", line 1261, in regtok_delete
    return self.query("delete", "v1/registration_tokens/{token}",
  File "/root/code/synadm/synadm/api.py", line 114, in query
    urlpart = urlpart.format(*args, **kwargs)
KeyError: 'token'
```

The rest of the commands seem to  work.

andreiz:
```
# git describe
v0.41.2-11-g11c7db7
```

andreiz:
`$ synadm regtok delete "mytoken"

Nothing out of the ordinary. 

The token seems to persist in the `context.params` at least until here:

`  File "/root/code/synadm/.venv/lib/python3.10/site-packages/click/core.py", line 1658`

[...]

PR #111, opened 2023-06-13
Fix regtok delete and others

JacksonChen666:
Fixes: #110

> Issue: `query` function accepts `token` as an arg, it doesn't go into `**kwargs` so `KeyError`.
> 
> May need to review other things that use conflicting keywords.

https://github.com/JOJ0/synadm/issues/110#issuecomment-1590032158

This PR fixes the conflicting keywords issue. I haven't found anymore of the same issue (I checked).

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -1148,8 +1148,8 @@ class SynapseAdmin(ApiRequest):
                 an exception occured. See Synapse admin API docs for details.
 
         """
-        result = self.query("get", "v1/registration_tokens/{token}",
-                            token=token)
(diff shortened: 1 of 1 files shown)


JOJ0 (approved):
LGTM. Thanks for the fast reaction and fix!
Also good idea to leave that comment about t instead of token kwarg. I think it's fine to state it on just one of the regtok_* api methods and not on all of them. That would be overkill. We/Contributors will find it when we need it. Please go ahead with merge and release when it's appropriate for you. Thanks again!

PR #113, opened 2023-07-26
Implement room delete v2 API

### LINKED ISSUE #49: Support v2 of the room delete API
and make sure pre-1.48 Synapse is still using delete API v1: https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#delete-room-api

Also the new v2 API changed some behaviour:
```
block - Optional. If set to true, this room will be added to a blocking list, preventing future attempts to join the room. Rooms can be blocked even if they're not yet known to the homeserver (only with Version 1 of the API). Defaults to false.
```

### LINKED ISSUE #50: Support new API's to query room delete status
Add one new subcommand supporting these two API's:

https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#query-by-room_id
https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#query-by-delete_id

JacksonChen666:
fixes #49
fixes #50

### DIFF
--- a/synadm/cli/room.py
+++ b/synadm/cli/room.py
@@ -19,6 +19,7 @@
 """
 
 import click
+from click_option_group import RequiredMutuallyExclusiveOptionGroup, optgroup
 
 from synadm import cli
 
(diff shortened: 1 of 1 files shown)


JOJ0:
merge to master plz. I'm afk and cant find the option to cgange on mobile.

synadm/cli/room.py
```diff
@@ -245,10 +246,13 @@ def members(helper, room_id):
     "--no-purge", is_flag=True, default=False, show_default=True,
     help="""Prevent removing of all traces of the room from your
     database.""")
+@click.option(
+    "--v1", is_flag=True, default=False, show_default=True,
+    help="""Use version 1 of the room delete API instead of version 2""")
```
JOJ0, line 251: I like this idea very much. I always had some auto-magic in my mind that detects when Synapse is old enough to use v1, otherwise v2.

I think having it "selectable" by the user/admin like this is a much easier and less errorprone solution. In the end a Matrix admin should have the right to choose what they want to do. Too much magic is not good anyway.

Perfect solution!

PR #117, opened 2023-08-05
Fix type checks in ApiHelper.load()

JOJ0:
Use isinstance() instead of type() as the current flake8 version suggests.


JOJ0:
I merged this now since is a tiny chance and I'm sure it's the right thing to do :-)

PR #118, opened 2023-08-05
Use requests.request instead of getting attrs to make requests

JacksonChen666:
Getting attributes is kind of hacky and not really that readable (or easily understood), so requests.request was used because it accepts HTTP methods.

(yeah I left the commit hanging around for a good few days)

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -131,8 +131,8 @@ class ApiRequest:
             override_verify = verify
 
         try:
-            resp = getattr(requests, method)(
-                url, headers=self.headers, timeout=self.timeout,
+            resp = requests.request(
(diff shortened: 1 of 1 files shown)


JOJ0:
@JacksonChen666 please pull --rebase from master, it will fix your flake8 issue. :+1:

JOJ0:
I triggered a rerun with debug loggin enabled of the linter gh-actions thing. let's see...

JOJ0:
Still same error. It's odd since it points to exactly those lines I fixed recently.

My suggestion would be: Go ahead and merge it.

I loosely assume this is some caching issue w gh-actions or the like.

If it happens in master again I'll look into it.

PR #123, opened 2023-09-02
admin filter for user listing

### LINKED ISSUE #122: Admin user filter in user list API
Another tiny new feature we might want to support:

Add an admins query parameter to the [List Accounts](https://matrix-org.github.io/synapse/v1.91/admin_api/user_admin_api.html#list-accounts) [admin API](https://matrix-org.github.io/synapse/v1.91/usage/administration/admin_api/index.html), to include only admins or to exclude admins in user queries. (https://github.com/matrix-org/synapse/issues/16114)

JacksonChen666:
the admin filter is a tri-state, meaning it can filter for admins, filter for non-admins, and not filter at all.

fixes #122

~~NOTE: It is a draft PR because I am looking for a cleaner solution of "the user said no based on this flag here" rather than the [current manual solution](https://github.com/JOJ0/synadm/pull/123/files#diff-a41082dbe9c56e01f08c3854a53edc1f26bae33f9de966cd4ed8e228fc2bf621R84-R88) of checking. If that's not possible, then this is pretty much complete.~~ found a cleaner solution

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -415,7 +415,7 @@ class SynapseAdmin(ApiRequest):
         self.user = user
 
     def user_list(self, _from, _limit, _guests, _deactivated,
-                  _name, _user_id):
+                  _name, _user_id, _admin):
         """List and search users
 
         Args:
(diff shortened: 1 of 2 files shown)


a-0-dev:
This was a breaking change for us (using `synadm` for scripting). The `user_list_paginate` function uses `user_list`, but the call of `user_list` was not updated to include some setting for the admin filter, so calling `user_list_paginate` inherently fails.

Correct me if I'm wrong, but to me it looks like this may be the problem, so potenially just fixing api.py:471 or extending `user_list_paginate`'s signature with the admin filter would be sufficient. If doing the latter, please consider putting a default value for the admin filter so there's no permanent breaking change for others using the function :)

Thanks for your work on synadm, we highly appreciate it!

PR #125, opened 2023-09-03
implement user (un)locking

### LINKED ISSUE #121: new matrix user admin function: lock
There is a bright new function added to matrix: user locking:

https://github.com/matrix-org/synapse/pull/15870/files

It would be awesome to have it available in synadm as well.

JacksonChen666:
fixes #121

did not go with `synadm user lock`, since it seemed like it was gonna be extra work (due to modify api being one big monolithic thing for many things) and just adding on to `synadm user modify` was actually easier.

### DIFF
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -426,7 +426,7 @@ class UserModifyOptionGroup(RequiredAnyOptionGroup):
 @optgroup.option(
     "--lock/--unlock", "-l/-L", default=None, show_default=False,
     help="""Whether to lock or unlock the account, preventing or allowing
-    logins respectively.""")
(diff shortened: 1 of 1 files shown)


JOJ0 (approved):
I tested on my prod server with a Synapse version that didn't have this admin-api-feature yet: It tries to set locking but it gets ignored without throwing an error. Alsu the "current settings" does not include locking yet (we pull that via `synadm user details`.

I updated my Synapse to latest, now I see `locked: false` in the "current settings" list, and it can be successfully locked and unlocked.

LTGM, merge it please! :-)

ser:
Thank you very much! Appreciate.

ser:
So maybe it's worth of new release? :)

PR #126, opened 2023-09-04
Rewriting the user modify command

### LINKED ISSUE #100: User modify API is poorly defined and only half works

JacksonChen666:
Currently a placeholder. Only about the structure of commands (and maybe the parameters) until agreed.

Initial idea: https://github.com/JOJ0/synadm/issues/100#issuecomment-1704264088

Resolves #100

[Placeholder/existing commands checklist](https://github.com/JOJ0/synadm/pull/126#issuecomment-1725298794)

(Commits history will 100% get rewritten after the code is finished)

---

Progress checklist:
- [x] Placeholders (see [checklist](https://github.com/JOJ0/synadm/pull/126#issuecomment-1725298794) and https://github.com/JOJ0/synadm/pull/126#issuecomment-1733108725)
- [x] Agreed placeholders ([agreed](https://github.com/JOJ0/synadm/pull/126#issuecomment-1742535566))
- [x] Implementation
- [ ] TODOs (deprecating, check `synadm user reactivate` with `--batch`)
- [ ] Testing everything new (including deprecation warning)
- [ ] Make README up to date with these changes

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -605,6 +605,7 @@ class SynapseAdmin(ApiRequest):
 
         Threepid should be passed as a tuple in a tuple
         """
+        # TODO: deprecate
         data = {}
         if password:
             data.update({"password": password})
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -505,6 +505,11 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
         click.echo("Abort.")
 
 
+# for these placeholders, function similarly to the modify command, but in
+# it's own command.
+# TODO: find 3pid modify command
+
+
 @user.command()
 @click.argument("user_id", type=str)
 @click.argument("user_type", type=str)


synadm/api.py
```diff
@@ -600,6 +600,7 @@ def user_modify(self, user_id, password, display_name, threepid,
 
         Threepid should be passed as a tuple in a tuple
         """
+        # TODO: deprecate
```
JOJ0, line 603: I agree we should deprecate `synadm user modify` within this PR, or actually I'd like to achieve this: As soon as everything from `user modify` has its own command _or_ the user should be warned - best with a literal log.warning.

But I'm not sure anymore wether we'd like to keep some parts of `user modify` as-is and not invent new commands. Can't recall what was the outcome of that discussion, please help me remember! :-)

If we _would_ keep parts of `modify` as-is and do not plan to change, then probably it would be better that exactly when these options are used, a log.warning is spit out. Not sure though what's best....

synadm/cli/user.py
```diff
@@ -500,6 +500,80 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
         click.echo("Abort.")
 
 
+# for these placeholders, function similarly to the modify command, but in
+# it's own command.
+# TODO: find 3pid modify command
```
JOJ0, line 505: We have a 3pid command already, but it's for displaying only: https://synadm.readthedocs.io/en/latest/synadm.cli.user.html#synadm-user-3pid

That I guess is what you are thinking about? Where should we put / and how should we name the modify command for 3pid modify?

Currently 3pid modify is included with the original `user modify` command: https://synadm.readthedocs.io/en/latest/synadm.cli.user.html#cmdoption-synadm-user-modify-t

synadm/cli/user.py
```diff
@@ -500,6 +500,80 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
...
+    # TODO
+    pass
+
+
+@user.command()
+@click.argument("user_id", type=str)
+@click.pass_obj
+def reactivate(helper, user_id):
+    # TODO
+    # TODO: does this require a password? merge into password? user
```
JOJ0, line 537: It definitely does require a password.

Look for this phrase "Note: the password field must also be set if both of the following are true:" around here: https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#create-or-modify-account

synadm/cli/user.py
```diff
@@ -500,6 +500,80 @@ def modify(ctx, helper, user_id, password, password_prompt, display_name,
...
+    pass
+
+
+@user.command()
+@click.argument("user_id", type=str)
+@click.pass_obj
+def admin(helper, user_id):
+    # TODO
+    # removal of admin privileges is in here as well, which isn't consistent
+    # with lock and unlock, but what's the word for un-admin?
```
JOJ0, line 573: We could go the "two word route" here as well:

Would be nicely sorted in `synadm user -h`:
`synadm user admin-grant`
`synadm user admin-revoke`

Would be kind of actual phrases:
`synadm user grant-admin`
`synadm user revoke-admin`

JOJ0 (commented):
Conclusive thoughts: If there are verbs in a command they are self-descriptive. Eg. `user deactivate`

If there are not they are not 100% obvious what they do (display or modify). Eg. `user display-name`

Not sure if I'm nitpicking here too much or if the priority should be "shorter" names. Eg. `set-display-name` is rather tedious to type and might be a bad choice...

synadm/cli/user.py
JOJ0: I basically agree but on the other hand: why not borrow the ideas we have over at the `user password` command?

I think we have both: stdin and cli arg

I still think that securitywise it's an admin's choice whether they want to eg. modify 10 testusers on a playground server - scripted! Or they'd like to securely modify s real user's account on the cli.

It's an old unix philosophy: Know what you are doing if you are root!

JOJ0:
> checklist of what's been planned/already done so far:
> 
> * [x]  password
> * [x]  display name
> * [x]  avatar URL
> * [ ]  changing 3pid
> * [ ]  `external_ids` (not part of `synadm user modify`?)
> * [x]  admin grant
> * [x]  admin revoke
> * [x]  deactivation
> * [x]  reactivation
> * [x]  locking
> * [x]  unlocking
> * [x]  user types

Nice summary! Why not move it to the initial description of this PR? Easier to find and also for future reference what's still missing/postponed (as I just read in your subsequent post)

synadm/cli/user.py
JOJ0: > Those arguments will probably also apply to the password command, if it doesn't have such (and it doesn't seem like the password command has such options).

Oh my bad. When I talked about "borrowing ideas of `user password` command" I probably ment the `user modify` command - We have `--password` and `--password-prompt` there.

As we had it so often already with Click/synadm, copying boilerplate code is the quick way but it would be nice if we  could find a way to prepare these options once and reuse them in several commands. That also applies to any sanity-checking code we'd require on the UI end.

JOJ0:
> > Conclusive thoughts: If there are verbs in a command they are self-descriptive. Eg. `user deactivate`
> > If there are not they are not 100% obvious what they do (display or modify). Eg. `user display-name`
> > Not sure if I'm nitpicking here too much or if the priority should be "shorter" names. Eg. `set-display-name` is rather tedious to type and might be a bad choice...
> 
> It should be with a `set` if it's going to set it. I didn't have plans to make it retrieve the display name and display it. Maybe a `--set` flag which can set a display name could work, while it could be handy to also see the display name (get by default, which will use the user query API).

Good question what's best....

I just had a look on your work so far in reality/--help:

```
  admin-grant
  admin-revoke
   ...
  set-display-name     Set the display name of a user.
  set-profile-picture  Set a profile picture for a user by the MXC URI.
  set-user-type
  ...
```

It seems there is no one-way of doing it right - verb in front, verb on the end of the command. Too long, too short command names....I'm not sure where to go actually....

One thing I'm wondering though: `--help` as the go-to move for most people (I hope! ;-)) - If we would find a way to be able to manually sort how the options are presented it would be less of a headache for the naming....

let me check, I think I researched this already ages agog and by default Click doesn't have it.....

JOJ0:
Ah ok, not sure how you ment that. You will move the checklist up into the description once it's finished?

Anyway, all good, work away! :-)

JOJ0:
Please go ahead with implementation and just use what you suggested / we have so far! Yes we could think out for ages on how commands should be named :-) Let's see when implementation is finished whether we'd like to change them.

synadm/cli/user.py
JOJ0: Regarding reusing of options or even a couple of options, I think I found a nice way we could try. Creating a decorator that applies a group of options. Have a look at the very bottom of this post: https://stackoverflow.com/a/66743881/10207149

What do you think?

JOJ0:
> Just realized, user modify API can also create users. Not sure how we'll deal with that, we might have to keep the user modify command around and possibly also update it with new stuff.

Oh, I'm sorry I didn't mention that, I was just assuming you are aware of that. I also didn't realise/think hard enough that it could be a problem with all the "redesign"..... :-/ Sorry my bad!

synadm/cli/user.py
JOJ0: > > * `--prompt` (default, mutually exclusive of `--no-prompt`)
> > * `--no-prompt` (mutually exclusive of `--prompt`, leave errors up to synapse)
> > * `--password` ([likely insecure](https://unix.stackexchange.com/questions/439497/is-there-a-way-to-pass-sensitive-data-in-bash-using-a-prompt-for-any-command) but useful for automation in testing environments, takes priority over stdin options)
> 
> Exactly, something like that!
> 
> > Those arguments will probably also apply to the password command, if it doesn't have such (and it doesn't seem like the password command has such options).
> 
> Oh my bad. When I talked about "borrowing ideas of `user password` command" I probably ment the `user modify` command - We have `--password` and `--password-prompt` there.
> 
> As we had it so often already with Click/synadm, copying boilerplate code is the quick way but it would be nice if we could find a way to prepare these options once and reuse them in several commands. That also applies to any sanity-checking code we'd require on the UI end.

Regarding password options streamlining, I would love if we could make use of the _common module I introduced here: https://github.com/JOJ0/synadm/pull/129/commits/79d0ea40eaf0e8101a3cca00cb88c73b1528a64d

We could put all the password related option into that module and reuse them with several password-related commands.

JOJ0: Sure, certainly I ment once it's merged ;-)

JOJ0: And yes: I agree we should also make use of that `click.password_option`

For some reason I did not use it in my intial implementation of `user password` - there might be a reason for it but it could also be that I just didn't read the docs carfully enough (it's been around since Click 6.x at least or even since always ;-). I used a confirmation prompt thing though :-)

```
$ synadm user password testuser3
Password:
Repeat for confirmation:
{}
```
https://github.com/JOJ0/synadm/blob/be77cef38e6889a37ec1b2acb9d28f2e58f83795/synadm/cli/user.py#L286-L288

We do also have use-cases were this option might not be enough....for example when handling with tokens there is more to it than either by prompt or by passing a cli arg... probably off-topic here....just mentioning..

JOJ0:
> > Just realized, user modify API can also create users. Not sure how we'll deal with that, we might have to keep the user modify command around and possibly also update it with new stuff.
> 
> Oh, I'm sorry I didn't mention that, I was just assuming you are aware of that. I also didn't realise/think hard enough that it could be a problem with all the "redesign"..... :-/ Sorry my bad!

To wrap this up and partly spice with some new ideas. I was thinking about all this a little:

The current `api.user_modify()` becomes the `user_create()` function - we need to keep it anyway as you mentioned. It _only_ is used for create, that should spare us at least a little headache because some things are just not relevant when creating). For tri-state options, we could even think if `enums` could help? Or we we keep the current `None`, `True`, `False` business....maybe that's enough. Don't kill me yet, I didn't think through each and every thing we want to set. 

Modifying is not possible anymore with this method! The newly designed `user set... whatever` commands take care of that, each with their own api methods, just as you designed and drafted it already.

Maybe we can reuse _ at least some_ of the cli-options by putting them into the `cli.common` module? Not sure! Often the wording in `--help` will be different, which makes reusing not possible.

PR #128, opened 2023-09-14
Fix existing calls to user_list API breaking

### LINKED ISSUE #127: User list function breaking change
> This was a breaking change for us (using `synadm` for scripting). The `user_list_paginate` function uses `user_list`, but the call of `user_list` was not updated to include some setting for the admin filter, so calling `user_list_paginate` inherently fails.
> 
> Correct me if I'm wrong, but to me it looks like this may be the problem, so potenially just fixing api.py:471 or extending `user_list_paginate`'s signature with the admin filter would be sufficient. If doing the latter, please consider putting a default value for the admin filter so there's no permanent breaking change for others using the function :)
> 
> Thanks for your work on synadm, we highly appreciate it!

_Originally posted by @a-0-dev in https://github.com/JOJ0/synadm/issues/123#issuecomment-1719890583_

JacksonChen666:
fixes #127

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -415,7 +415,7 @@ class SynapseAdmin(ApiRequest):
         self.user = user
 
     def user_list(self, _from, _limit, _guests, _deactivated,
-                  _name, _user_id, _admin):
+                  _name, _user_id, _admin=None):
         """List and search users
 
         Args:


synadm/cli/notice.py
```diff
@@ -135,7 +135,8 @@ def confirm_prompt():
         formatted_content = formatted if formatted else plain_content
 
     if regex:
-        if "users" not in helper.api.user_list(0, 100, True, False, "", ""):
+        if "users" not in helper.api.user_list(0, 100, True, False, "", "",
+                                               None):
```
JOJ0, line 139: There is another call of user_list in line 97 in this file. I think that also requires this additional argument. Thanks!

JOJ0 (changes requested):

JOJ0 (approved):
LGTM. Thanks!

PR #129, opened 2023-10-02
Add `raw` command, a new module for common CLI options, and improve `matrix raw` help

### LINKED ISSUE #11: Add ability to call custom GET/POST command via synadm script
Will be good to use same `synadm` tool not only for pre-defined API calls, but for custom too, for not to create custom scripts for each call.
For example, allow something like this:
```
synadm custom-api-call --type POST --path "/_synapse/admin/v1/media/ru-matrix.org/delete?before_ts=1602745639000" --body "{}"
```
where make default argument value for `type` as `GET`, `body` = `{}`.

This way will be much more comfortable, than composing full `curl` command in cli manually and remember your token each time.

Especially this is useful at now, when some of popular API calls are not implemented in build-in commands.

JOJ0:
Addressing 2 things:

- A new subcommand `synadm raw` that should, very similar to what we have in`synadm matrix raw`, be able to rudimentary issue custom requests (that don't have their own synadm commmand yet). Fixes #11
- Finally provide a way to a common problem in the synadm project: Often it's desired to share options or groups of options with multiple commands, even when they are in different "main commands" (live in different modules).
  - Introducing a new module `synadm.cli._common` containing such options and option groups.
  - The new `raw` command will make use of it.

### DIFF
--- a/doc/source/index_cli_reference.rst
+++ b/doc/source/index_cli_reference.rst
@@ -15,3 +15,4 @@ Command Line Reference
    synadm.cli.matrix
    synadm.cli.regtok
    synadm.cli.notice
+   synadm.cli.raw
--- a/doc/source/synadm.cli.raw.rst
+++ b/doc/source/synadm.cli.raw.rst
@@ -1,0 +1,6 @@
+Raw
+====
+
+.. click:: synadm.cli.raw:raw_request_cmd
+   :prog: synadm raw
+   :nested: full
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -1366,3 +1366,17 @@ class SynapseAdmin(ApiRequest):
         else:
             data["user_id"] = receivers
             return [self.query("post", "v1/send_server_notice", data=data)]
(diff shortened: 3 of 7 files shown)


JOJ0:
Thanks for a quick review. Not so much about the technicalities, but about the many fixes/improvements/clarifications I tried to squeeze into --help of both, the existing `matrix raw`, as well as the `raw` command.

I tested the feature and since it's a simpler version of `matrix raw` which is tried and tested there shouldn't be much to worry.

The behaviour of the both is streamlined as you will see when comparing their `--help`

JOJ0:
For maximum review convenience, here screenshots how rendered docs look like:

Screenshot 2023-10-13 at 12 35 54


Screenshot 2023-10-13 at 12 35 37

JOJ0:
@JacksonChen666 I decided that this is most probably good and spare you the testing/reviewing. I think it's find. Let's get it out there. I'll even release it shortly....

PR #136, opened 2023-11-13
User modify command threepid handling fixes

### LINKED ISSUE #135: Removing of threepid is not supported
From what I understand of [the docs](https://synadm.readthedocs.io/en/latest/synadm.cli.user.html#cmdoption-synadm-user-modify-t) and my tests, this option does not allow to remove/override the existing 3pids, only to add more:

> `-t, --threepid <threepid>`
>
> Add a third-party identifier. [...]

_Originally posted by @n-peugnet in https://github.com/JOJ0/synadm/issues/100#issuecomment-1807097561_

JOJ0:
Fixes #135

Improve the `--threepid`option of the `user modify` command.

- Support removing **all** threepids by passing `--clear-threepids`
  Note: If a user has the idea/follows the logic that resetting according to the API docs (https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#create-or-modify-account) should also work by setting it to "empty", ie. `--threepid '' ''`, that will also work, but it is not documented in `--help`!
- Extend and clarify `--help` for this option.
- Fix an issue where passing multiple `--threepid` options, sets only the last of the argument pairs the user had provided on the cli (ignoring all others).

### DIFF
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -394,14 +394,15 @@ class UserModifyOptionGroup(RequiredAnyOptionGroup):
     help="Set display name. defaults to the value of user_id")
 @optgroup.option(
     "--threepid", "-t", type=str, multiple=True, nargs=2,
-    help="""Add a third-party identifier (email address or phone number).
-    Threepids are used for several things: For use when logging in, as an
-    alternative to the user id; in the case of email, as an alternative contact
-    to help with account recovery; as well as to receive notifications of
-    missed messages. This option requires two arguments: `medium value` (eg.
-    `--threepid email <user@example.org>`). To configure multiple entries for a
-    user, this option can be passed multiple times. To clear _all_ threepids,
-    pass it _once_ while setting two empty strings: '' ''.""")
+    help="""Set a third-party identifier (email address or phone number). Pass
+    two arguments: `medium value` (eg. `--threepid email <user@example.org>`).
+    This option can be passed multiple times, which allows setting multiple
+    entries for a user. When modifying existing users, all threepids are
+    replaced by what's passed in all given `--threepid` options. Threepids are
+    used for several things: For use when logging in, as an alternative to the
+    user id; in the case of email, as an alternative contact to help with
+    account recovery, as well as to receive notifications of missed
+    messages.""")
 @optgroup.option(
     "--avatar-url", "-v", type=str,
     help="""Set avatar URL. Must be a MXC URI
--- a/synadm/cli/user.py
+++ b/synadm/cli/user.py
@@ -394,14 +394,14 @@ class UserModifyOptionGroup(RequiredAnyOptionGroup):
     help="Set display name. defaults to the value of user_id")
 @optgroup.option(
     "--threepid", "-t", type=str, multiple=True, nargs=2,
-    help="""Add a third-party identifier. This can be an email address or a
-    phone number. Threepids are used for several things: For use when
-    logging in, as an alternative to the user id. In the case of email, as
-    an alternative contact to help with account recovery, as well as
-    to receive notifications of missed messages. Format: medium
-    value (eg. --threepid email <user@example.org>). This option can also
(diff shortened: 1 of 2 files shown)


JOJ0:
One more uglyness remains with this quick and dirty implementation. If the user passes some valid arg-pairs alongside one empty-string-pair, the sanity check still complains:

```
User account settings to be modified:
threepid: email abc@123.com
threepid: email def@456.com
threepid:
WARNING  is probably not a supported medium type. Threepid medium types according to the current matrix spec are: email, msisdn.
Are you sure you want to modify user? (y/N): y
INFO  Querying put on https://peek-a-boo.at/_synapse/admin/v2/users/%40testuser1%3Apeek-a-boo.at
errcode: M_INVALID_PARAM
error: ''''' is not a valid value for ''medium'''
```

Is that bad? Or ist that actually good (because bad arguments where passed anyway)?.

JOJ0:
@JacksonChen666 I know there is some rather dirty fixes in this PR. Some nitpicking on my help texts appreciated! Or maybe you have some better ideas? Or we decide that it is good enough for now?

@n-peugnet if you have a minute: Can you install from this branch and test it out in terms of "usability"?

synadm/cli/user.py
```diff
@@ -394,14 +394,14 @@ def name_extra(self):
...
     "--threepid", "-t", type=str, multiple=True, nargs=2,
-    help="""Add a third-party identifier. This can be an email address or a
-    phone number. Threepids are used for several things: For use when
-    logging in, as an alternative to the user id. In the case of email, as
-    an alternative contact to help with account recovery, as well as
-    to receive notifications of missed messages. Format: medium
-    value (eg. --threepid email <user@example.org>). This option can also
-    be stated multiple times, i.e. a user can have multiple threepids
-    configured.""")
+    help="""Add a third-party identifier (email address or phone number).
```
n-peugnet, line 397: So does it add 3pids or replace them in the end?

JOJ0: Ah yeah, thanks for that catch! It relaces for existing accounts. That was the reason I was rewriting this help text in the first place. The problem with wording right now is that this user modify command is used to create as well as alter user accounts. Suggestions how to write this concisely?

Set

n-peugnet:
> @n-peugnet if you have a minute: Can you install from this branch and test it out in terms of "usability"?

I just tested it. It works as expected, but it is a little bit strange to use. Maybe a `--clear-threepids` option would be clearer. For now I can live with the `--threepid '' ''` option.

synadm/cli/user.py
```diff
@@ -394,14 +394,14 @@ def name_extra(self):
...
+    Threepids are used for several things: For use when logging in, as an
+    alternative to the user id; in the case of email, as an alternative contact
+    to help with account recovery; as well as to receive notifications of
```
JOJ0, line 400: should be comma not semicolon!

JOJ0:
> > @n-peugnet if you have a minute: Can you install from this branch and test it out in terms of "usability"?
> 
> 
> 
> I just tested it. It works as expected, but it is a little bit strange to use. Maybe a `--clear-threepids` option would be clearer. For now I can live with the `--threepid '' ''` option.

Yes I also thought about a separate clear threepid option. The thing with that is that it's a weird option to have when creating an account. Well, actually, it would just be ignored in that case.

All these weirdnesses are some of the reasons for the redesign  approaches you already noticed ;-)

JOJ0:
I finally decided to add a separate option `--clear-threepids`, it doesn't make our code better readable (duh!) but at least it helps users. If you have another few minutes, some test runs are appreciated. Otherwise I'll merge and release this in a couple of days.

I also gave the help texts another rewrite and reordering. Hope I got everything to be told into it finally.

Also some fixes in wording of the "user interaction" messages. The command always stated that "something is modified" even though it could also be just freshly "created"

Anyway there is room for improvement / redesign in all this. We know...we know...known issue... :-)

PR #146, opened 2024-07-01
README rewrite (v2)

JacksonChen666:
Reduce down the size of the README.md file so that users don't miss tiny things like the PyPI installation method, etc.

This has been sitting for a while, may be a bit messy right now. Don't merge until further notice or cleanup.

([Self reference for things to keep/yeet (check before marking as ready)](https://matrix.to/#/%21CnFMflNNLilExvnfRs%3Apeek-a-boo.at/%24yDKJG3mJb87j5rL-1E6A674Xt_NSAR0WdpldotVrJAE?via=peek-a-boo.at&via=jacksonchen666.com&via=matrix.ascurius.com), also the big checklist)

([Previous version of README rewrite](https://github.com/JOJ0/synadm/pull/144))

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ To install synadm from PyPI, run the following command:
 
 `pip3 install synadm`
 
-Other installation methods: [Install via git](TODO)
+Other installation methods: [Install via Git](https://github.com/JOJ0/synadm/blob/master/CONTRIBUTING.md#getting-the-source--installing)
 
 ## Update
 


README.md
```diff
@@ -283,22 +209,18 @@ pip install .
   * [x] `regtok update <registration token>`
   * [x] `regtok delete <registration token>`
```
JOJ0, line 210: I suggest you kick out the entire list here. Sooner or later and not defined in what form exactly, I suggest here we just link to the Sphinx docs page that holds a feature list / implementation status list

JOJ0 (approved):
I carefully reviewed all the current changes in this draft via the rich-diff and I think this is ready to merge. I've added a tiny suggestion to complete a TODO. I think this is a good first iteration of a shorter and more concise README. We can always further shorten in future PR's. Good enough for now IMHO. What do you think? Merge?

Issue #133, opened 2023-10-14
GitHub Actions pipeline "release" throwing deprecations/warnings


JOJ0:
Several deprecations and warnings. I'll try to look into all that but on the other hand might as well even think about implementing an even more automated releasing approach....just thinking out loud....let's discuss via Matrix soonish...

![Screenshot 2023-10-14 at 12 52 48](https://github.com/JOJ0/synadm/assets/2733783/21da39a4-9937-4ac8-8225-1ea70bbd4483)

JOJ0:
Hi, I had started with fixing things in a branch back in October, but forgot about it. Thanks for the reminder and most of all thanks for all the details. I opened a draft pr and will incorporate those suggestions into it.

PR #154, opened 2024-09-10
Fix GitHub Actions deprecations and warnings

JOJ0:


JOJ0:
Sorry @JacksonChen666, not really ready for review. Just desperately trying to trigger the release workflow. As I understood it, it should be sufficient to push a tag named 'v*' to trigger it. But it doesn't seem to work...Hmm 🤔

JOJ0:
Ok got it, `retag.sh` helped to remind me on how to trigger/test all this.

Now it's ready to review. Note that in this case the failed release workflow is a good thing. It tells us that the workflow is running as planned an my test-tag-push and at least this "early exit" on version mismatch is doing its job correctly:

Screenshot 2024-09-10 at 09 31 15

(Now deleted test tag v0.46.2 again)

As far as I can see all the deprecations as documented in the screenshot in #133 (headline "Annotations") are gone, but please double check!

JOJ0:
Thanks for testing! Seems fine!

PR #158, opened 2024-10-09
Docs theme change

### LINKED ISSUE #156: Modernize the look of our generated docs
The readthedocs theme we are using, in my opinion, looks and feels outdated. I'm currently reading a lot of data science and machine learning related documentation. The PyData theme for Sphinx seems to be very popular and I have to admit that I like it very much. Have a look at these random docs links which use that theme :

- https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.columns.html
- https://matplotlib.org/stable/gallery/style_sheets/style_sheets_reference.html
- https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html#sklearn.cluster.KMeans

BTW I also would have experience in implementing the PyData theme, since I did it for the Beets project already (https://beets.rtfd.io).

Nevertheless I'd like to start a discussion on what theme would be best for the synadm project. Some more suggestions I like are:

- The "Book" theme: https://sphinx-themes.org/sample-sites/sphinx-book-theme/ - pretty similar to PyData but feels slightly more minimalistic. Actually I prefer it over PyData at the moment!
- The "Insegel" theme: https://sphinx-themes.org/sample-sites/insegel/
- The "Awesome" theme: https://sphinx-themes.org/sample-sites/sphinxawesome-theme/

Here's an overview of (all?) available Sphinx themes - but be aware of that not every theme has the same features when it comes to usability: https://sphinx-themes.org/

JOJ0:
Resolves #156

### DIFF
--- a/doc/source/_static/synadm.css
+++ b/doc/source/_static/synadm.css
@@ -1,0 +1,14 @@
+/*
+html[data-theme="light"] {
+    --pst-color-secondary: #a23632;
+}
+html[data-theme="light"] {
+    --pst-color-inline-code: #a23632;
+}
+*/
+table tr:hover td {
+    background-color: var(--bs-gray-300);
+}
+
+/* pydata teal (primary): #126A7E */
+/* pydata violet (secondary): #7D0E70 */
--- a/doc/source/_static/synadm.css
+++ b/doc/source/_static/synadm.css
@@ -15,6 +15,10 @@ table {
 }
 table,thead,tbody {width:98%}
 */
+html {
+    --pst-font-size-base: 14px;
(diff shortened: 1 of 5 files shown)


JOJ0:
Thanks for removing the furo theme adjustments. The width definitely needs to go. About the width, we could keep it, it's not wrong to set it and other themes have center as their default, see: https://github.com/sphinx-doc/sphinx/issues/4550

Anyway I'm just noting it here in case we ever need it because pydata theme decides to follow that default at some point. We can merge like this, less clutter in features.rst is good.

Issue #155, opened 2024-09-14
Include CONTRIBUTING.md into the generated docs


JOJ0:
but at the same time keep it as a plain MD formatted file in the root of the git repo, which fits with the (quasi) standard we see in other git (github?) repos.

Motivation and thoughts: When I look at our generated docs at https://synadm.readthedocs.io (https://synadm.rtfd.io), it always feels like developers documentation is entirely missing. We are not even linking to it. In my opinion one well formatted documentation "entry point" is a goal. It would also support the current efforts to simplify README.md as much as possible. We can always link to the generated Sphinx docs from there. The same goes for the elsewhere discussed TROUBLESHOOTING.md, I suggest to also incorporate this into the generated docs and link to it from wherever we see fit.

PR #161, opened 2024-10-24
Include CONTRIBUTING.md into Sphinx generated docs

JOJ0:
Adresses #155


JOJ0:
This has been lying around for too long. There were only a couple of hyperlinks left to fix, which I did yesterday.

That only thing that's a odd now is that we have top-level table of contents on this page, which wouldn't be necessary. We need that ToC in the "standalone version" though (people viewing CONTRIBUTING.md in the repo directly). Anyway, I don't see this as a problem.

I also checked for dead links with this handy command:

```
$ make linkcheck | grep 'contributing:.*broken'
```

FYI @JacksonChen666 no further review required. Let's kick this out to get even better synadm docs.

PR #163, opened 2024-12-01
hint that when synadm room join fails you can use synadm room make-admin

symphorien:


JOJ0 (approved):
LGTM. I wanted to suggest to further clarify where the make-admin ist to be found, but since we are staying in the context of the `synadm room` commands I think it's absolutely fine to simply state `make-admin`. Thanks for the help!

PR #165, opened 2025-01-17
Rework README install chapter and recommend `pipx`

JOJ0:
Slight rework of README install chapter and some tiny polishing elsewhere.


JOJ0:
Thanks for looking through and merge!

PR #168, opened 2025-01-24
Exit synadm early on fatal errors, refactor & fix `query()`

### LINKED ISSUE #164: `room power-levels` should fail gracefully when Synapse not reachable
```
$ synadm room power-levels
ERROR ConnectionError while querying Synapse: HTTPConnectionPool(host='localhost', port=8008): Max retries exceeded with url: /_synapse/admin/v1/rooms?from=0&limit=10 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x10405aa90>: Failed to establish a new connection: [Errno 61] Connection refused'))
...
...

  File "/Users/user/git/synadm/synadm/api.py", line 849, in room_power_levels
    for i, room in enumerate(rooms["rooms"]):
                             ~~~~~^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

JOJ0:
Addresses general error handling issues discussed in: #164

- What actually is a fatal error where it does not make sense to continue? I picked a few already. There might be more. Please suggest!
- Is it really a good idea to define `log_fatal_exit()` directly in this module as a regular function? Should it live inside `ApiRequest`? Or move it to a "utils-like" place? Will we use it outside this module too?
- Some refactoring along the way
- Add type hints (we can discuss if this is the right place and time....)
- Fix docstring
   - arguments done
   - return value: FIXME, isn't this entirely wrong? `resp.json()` returns lists or dicts. Or what does `json string` mean in this context? It's no string in reality....

### DIFF
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -31,10 +31,10 @@ https://matrix.org/docs/spec/#matrix-apis.
 import datetime
 import json
 import re
-from typing import Optional, Union, Dict, List, Any
+import traceback
 import urllib.parse
 from http.client import HTTPConnection
-import traceback
+from typing import Any, Dict, List, Optional, Union
 
 import requests
 
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -52,7 +52,7 @@ def log_fatal_exit(error, logger, message=None):
     if message is None:
         message = "synadm exited due to a fatal error."
 
-    # split by new lines that's already included
+    # format_exception() returns a list of strings. join it into a single string again. 
     logger.info("".join(traceback.format_exception(error)))
 
     logger.fatal(
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -36,7 +36,7 @@ import urllib.parse
 from http.client import HTTPConnection
 
 import requests
-from requests.exceptions import InvalidURL, MissingSchema, ConnectionError
+from requests.exceptions import ConnectionError
 
 
 def log_fatal_exit(error, logger, message=None):
@@ -161,10 +161,6 @@ class ApiRequest:
                                  f"{resp.status_code}")
             return resp.json()
         except ConnectionError as error:
-            log_fatal_exit(error, self.log)
-        except InvalidURL as error:
-            log_fatal_exit(error, self.log)
-        except MissingSchema as error:
             log_fatal_exit(error, self.log, "Connection error. Please check "
                            "that Synapse can be reached.")
         except Exception as error:
--- a/synadm/api.py
+++ b/synadm/api.py
@@ -30,10 +30,21 @@ https://matrix.org/docs/spec/#matrix-apis.
 
 import requests
 from http.client import HTTPConnection
+from requests.exceptions import InvalidURL, MissingSchema
 import datetime
 import json
 import urllib.parse
 import re
+
+
+def log_fatal_exit(error, logger):
+    """Log a fatal error and exit synadm."""
+    logger.fatal(
+        "%s: %s.\nsynadm exited due to a fatal error.",
+        type(error).__name__,
+        error,
+    )
+    raise SystemExit(1) from error
 
 
 class ApiRequest:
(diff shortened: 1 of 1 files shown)


JacksonChen666 (commented):
Looked at it, made some commits. Hope to get some feedback on some of my commits, if any.

synadm/api.py
```diff
@@ -28,12 +28,39 @@
...
 import json
-import urllib.parse
 import re
+from typing import Optional, Union, Dict, List, Any
+import urllib.parse
+from http.client import HTTPConnection
+import traceback
+
+import requests
+from requests.exceptions import ConnectionError
```
JOJ0, line 40: My linter keeps telling me that this is not required to import since it is built-in. The damn thing is wrong though and you are right. This must be imported otherwise the exception is not catched. (Damnit `pylint` I was counting on you! 🤣 )

synadm/api.py
```diff
@@ -28,12 +28,39 @@
...
+        error: A Python exception.
+        logger: A Python logger, with info and fatal methods
+        message: Message to use instead of "synadm exited due to a fatal
+            error."
+    """
+    if message is None:
+        message = "synadm exited due to a fatal error."
+
+    # split by new lines that's already included
+    logger.info("".join(traceback.format_exception(error)))
```
JOJ0, line 56: I had to read the docs to understand what's happening here and understand what you try to tell in the comment: https://docs.python.org/3/library/traceback.html#traceback.format_exception

synadm/api.py
```diff
@@ -28,12 +28,39 @@
...
+    Args:
+        error: A Python exception.
+        logger: A Python logger, with info and fatal methods
+        message: Message to use instead of "synadm exited due to a fatal
+            error."
+    """
+    if message is None:
+        message = "synadm exited due to a fatal error."
+
+    # split by new lines that's already included
```
JOJ0, line 55:
```suggestion
    # format_exception() returns a list of strings. join it into a single string again. 
```

Would this still make sense?

I would like this see how this looks like in action. Can you give me a hint on how to produce a traceback. I can't seem to find a way to trigger one. Thanks!

synadm/api.py
```diff
@@ -73,26 +100,36 @@ def __init__(self, log, user, token, base_url, path, timeout, debug,
...
+        method: str,
+        urlpart: str,
+        *args: Any,
+        params: Optional[Dict[str, Any]] = None,
+        data: Optional[Dict[str, Any]] = None,
+        token: Optional[str] = None,
+        base_url_override: Optional[bool] = False,
+        verify: Optional[bool] = True,
+        **kwargs: Dict[str, Any],
+    ) -> Optional[Union[Dict[str, Any], List[Dict[str, Any]], None]]:
```
JOJ0, line 114: This is pretty complicated and hard to read already. You wrote in the summary that we could actually receive a single "string" or number type without any wrapping dict or list? 

The current definition is telling that we get a list, a dict or None. Will this be accurate enough? Or should we even turn this into "returns anything or None"

JOJ0: Looking at this again. The `, None` near the end of the line is redundant, actually `Optional` already tells that `None` could also be returned. This really tells what I intended to tell. Still same question: Making it more complicated would make sense yout thing or is it good enough?
```suggestion
Optional[Union[Dict[str, Any], List[Dict[str, Any]]]]
```

synadm/api.py
JOJ0: I also found this article about itL https://stackoverflow.com/questions/1378889/how-to-format-traceback-objects-in-python

JOJ0: A minute after writing this, starting my work-day, it came to me "Wasn't this a logger.info(), of course I don't see it, it's in the log or with verbose on console. Sorry for the question. Got it :-)

synadm/api.py
```diff
@@ -52,7 +52,8 @@ def log_fatal_exit(error, logger, message=None):
     if message is None:
         message = "synadm exited due to a fatal error."
 
-    # format_exception() returns a list of strings. join it into a single string again. 
+    # format_exception() returns a list of strings (with new lines). join it
+    # into a single string again for readability.
```
JOJ0, line 56: perfect! thanks!

synadm/api.py
JOJ0: Yes, that's how we should do it: 4bb093d. Pylint was actually complaining that I overwrite the built-in with `as ConnectionError`. This is also fixed now.

JOJ0:
Good to go? @JacksonChen666

JOJ0:
> Good to go? @JacksonChen666

Ah moving to utils place......

JacksonChen666 (approved):

